PR #4, opened 2020-04-20
Fix ccache

FLGX06:
This seems to fix ccache for me. The speed of it then is just awesome.


Tk-Glitch:
That variable doesn't exist (anymore) in linux-tkg and makepkg only reads the `options` array, which would require `options=(${_makepkg_options[@]})` in place of the current options array, and as such that its value was set much earlier than what your proposition does. In its current form your PR has zero effect :frog:
Your results are probably due to your ccache getting populated while experimenting.

FLGX06:
I thought it's because ccache getting populated too, and apparently that's what happened. I just hoped it isn't that, but only changing scheduler is enough to make the whole code recompile, which is weird.

PR #18, opened 2020-06-11
linux5*-tkg: Fix finding config fragments when $_where is a symlink

JamiKettunen:
I personally prefer to symlink the folder containing the kernel `PKGBUILD`
among other files to somewhere else for quicker access and with that
setup `.myfrag` config fragments were not getting found earlier.


Tk-Glitch:
I suppose that works! Thank you 👍

PR #20, opened 2020-06-17
linux54/56/56-tkg: Fix MuQSS configuration comment

CodingKoopa:
This PR adds some missing quotes to `customization.cfg` for consistency, and adds `muqss` where there should be one, as linux57 does seem to support it. Cheers.


Tk-Glitch:
🐸 👍

PR #22, opened 2020-06-18
linux57/58-tkg: Add _noccache to default cfg.

CodingKoopa:
These lines seem to have been left behind. [`linux56-tkg`](https://github.com/Frogging-Family/linux-tkg/blob/14df411ac59ed7cef254a21732f27b904c336d99/linux56-tkg/customization.cfg#L26-L27) for reference.


Tk-Glitch:
Looks like it indeed. Thanks!

Issue #21, opened 2020-06-17
Build for Ubuntu


AdelKS:
Hello!

First of all, thank you for this nice repository with select patches for Linux gaming. I hope I am not annoying with this issue. 

I want to use your repository to build a custom kernel for my Ubuntu 20.04 with Fsync and march=native optimisations with GCC 10.1. Unfortunately I couldn't find a way to make `makepkg -si` work (it seems that it's part of Arch's `pacman`). For Ubuntu, I found [this wiki page](https://wiki.ubuntu.com/KernelTeam/GitKernelBuild) to build from the Linux git tree and it works awesomely, too simple to be honest. I expected more work.

A possibility to do what I want is that I only apply the patches in the folder `linux57-tkg-patches` after a `git checkout v5.7` in the linux tree, but I feel like I will be missing some tweaks from the other files on your repo, that I suppose will tweak the `.config` file. 

Do you have any better ideas on how I can use your repository to build an Ubuntu kernel (that makes .deb packages) ?

Thank you!

Adel

AdelKS:
I successfully compiled a fully working v5.7 kernel with your patches, I mainly followed the `PKGBUILD` script off by hand and chose what I wanted. I think I can derive a shell script that works on Ubuntu from that file, a very basic version first, would you be interested in integrating it in the repository ?

Tk-Glitch:
It was talked about in https://github.com/Tk-Glitch/PKGBUILDS/issues/539 (which is basically a duplicate) and is still on my to-do list (with very low priority, I have to admit). The whole nightmare with wine right now has been slowing my other works down.

My plan is to have a core script for source preparation that would then get used by the PKGBUILD and, as/if needed, other scripts (in a similar fashion to wine/proton-tkg) for portability sakes.

A Ubuntu-centric script would definitely be of help.

AdelKS:
Oh I haven't followed closely what's happening with wine, though I did clone it and to compiled the latest master version without thinking further than modding the `customisation.cfg` file. Good luck with all that !

> My plan is to have a core script for source preparation that would then get used by the PKGBUILD and, as/if needed, other scripts (in a similar fashion to wine/proton-tkg) for portability sakes.

[...]

PR #26, opened 2020-06-29
Add Ubuntu install script + general-ish .config&patching script

AdelKS:
Hello!

This is my current work after opening the issue #21. The implemented changes so far are the following : 
- I added the config option ``_distro`` in ``customization.cfg`` where the user should put "Arch", "Ubuntu" or "Fedora". This options is used for two reasons:
  - For the install script ``install.sh`` to know which distro it's working on, it will be important when installing since Fedora and Ubuntu are different
  - To isolate some code in ``_tkg_srcprep()`` that isn't useful for non-arch installs.
  - Extra: This option can be avoided by checking out ``/etc/os-release``, where the ``NAME`` variable contains the distro's name. Sounds rather appealing to me, but I didn't want to bother for now. What do you think of it ? That would make the option ``_distro`` useless. But would mean that Arch derivatives, Debian derivatives and Redhat derivatives should be recognized appropriately. 
- I added an ``install.sh`` script, for now it only downloads the linux kernel sources, patches them, and does the config starting off the distro's current one. 
  - Please try the install script on your machine, even if it's Arch based, lie to the script by telling it you're using Ubuntu for you to see what's happening, if you have any comments on how things are done.
- I made ``_tkg_srcprep()`` a bit more verbose, just so I can visually see when a patch starts and when it ends. Up to you if you want to keep it.
- I added ``make xconfig`` option.

I will work on the Ubuntu install and uninstall process. I will base it off a custom name, given by the user, the recognize the ``deb`` packages installed by the script and uninstall them. After that, I will tackle Fedora.

Also, please check that I haven't broken the PKGBUILD process.

Tell me what you think, I will keep you updated.


AdelKS:
Speaking of ``make xconfig``, when I reach that stage with my current setup I get the following warnings then the window launches:

> .config:6639: warning: override: reassigning to symbol CMDLINE
> .config:6652: warning: override: reassigning to symbol MATOM
> .config:6679: warning: override: HZ_500 changes choice state

If I save then quit, without doing any modifications, the generated fragments are the following:

> CONFIG_HZ=500
> # CONFIG_HZ_250 is not set
> # CONFIG_HZ_750 is not set
> # CONFIG_MNATIVE is not set
> CONFIG_SMT_NICE=y

I am not sure this is related with my script, but I am writing it if ever it's the case.

AdelKS:
@Tk-Glitch Can the function ``user_patcher()`` be moved to the ``prepare`` file ? So I can enable in in the non-Arch install script.

Tk-Glitch:
`user_patcher()` can indeed be moved to the prepare file. I don't think the config overrides are related to your script. Those are looking familiar and normal to me.

Due to health issues I won't be able to properly review your code until a few days. Sorry about that.

AdelKS:
> `user_patcher()` can indeed be moved to the prepare file. I don't think the config overrides are related to your script. Those are looking familiar and normal to me.

Got it! I will move it then.

> Due to health issues I won't be able to properly review your code until a few more days. Sorry about that.

Don't you even worry, I will keep on working on the script meanwhile, I hope it's nothing serious. get better soon ! 😄

AdelKS:
With the latest sub-versions of linux-5.7, 5.7.6 and 5.7.7 so far I have been getting these errors, they happen when the `.deb` files are being created, I can't find anything about it on the webs:

> error: include/uapi/linux/vt.h: leak CONFIG_NR_TTY_DEVICES to user-space
make[6]: *** [scripts/Makefile.headersinst:63: usr/include/linux/vt.h] Error 1
make[6]: *** Waiting for unfinished jobs....
  HDRINST usr/include/linux/kvm_para.h
make[5]: *** [Makefile:1227: headers] Error 2
make[4]: *** [scripts/Makefile.package:87: intdeb-pkg] Error 2
make[3]: *** [Makefile:1463: intdeb-pkg] Error 2
make[2]: *** [debian/rules:9: binary-arch] Error 2
dpkg-buildpackage: error: fakeroot -u debian/rules binary subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:77: deb-pkg] Error 2
make: *** [Makefile:1463: deb-pkg] Error 2

Any idea what the reason behind this can be ?

AdelKS:
Given I am having problems with the latest kernel sub-versions, I have found a very elegant way to switch between sub-versions and to reset the sources to their original state when re-running the install script : by using git !

I found a branch, `linux-5.7.y`, in https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git, that only contains the improvements to the linux-5.7 base kernel. Starting from there:
1. Shallow clone that branch up to 5.7.0
2. Use `git checkout v.5.7.x` between versions when needed
3. Use `git clean -f -d` to reset the source files to their original state.

What I win with this approach is that the hard drive is not over-used when resetting sources to their original state when re-running the install script. Because, for resetting, I was doing `rm -rf source-folder` then unpacking it again from the original tar archive.

AdelKS:
@Tk-Glitch The new patches you put seem to fail with linux 5.7.5 and under. And since I cannot install versions above that version, because of the error I am getting (two messages above), I restored the previous patch versions with an `.old` extension. Tell me if you have a problem against that.

The script finally can install the custom compiled Kernel, I only need to add commands to the install script. I am thinking of the following:
- `./install.sh prepare` to do all the source preparation before calling make
- `./install.sh build` to do the building
- `./install.sh install` to install, it will store the name of the installed custom kernel in the file `installed-kernels`
- `./install.sh uninstall` to uninstall one custom kernel, referenced in the `installed-kernels` file (the user will chose from a list of in a promp)
- `./install.sh clean` to do clean everything except `installed-kernels`, maybe add an option to exclude other files (for example the user's custom `customization.cfg`)

You can give a try to the install script, you will have to comment out the `dpkg -i` lines at the bottom

AdelKS:
Okay now there's a third version of the patches due to 5.7.8. I will figure out a nice naming scheme... ".old2" ? xD

Tk-Glitch:
Hello again AdelKS!

So, where to start ? :frog: 
Having multiple subversions around is undesirable. It frogs up version tracking on top of adding unnecessary fluff. Having multiple versions of the patches around for each subversion is not my intent either.

The issue dpkg seems to have with leaky `CONFIG_NR_TTY_DEVICES` reveals having an universal solution doesn't work too well. I can obviously get rid of the patch or move it to its own file to make it optional, but who knows what else might randomly trigger dpkg and when. I won't know until someone reports it, and tracking such distro-specific issues might become unmanageable for me pretty fast. That was one of my fears with this and having it happen already doesn't bode too well :sweat_smile: What do you think?

Making git clones, while offering more flexibility, breaks rebased patchsets unless you target the same tree as the patches, as you noticed. This defeats the purpose of having them in the first place in my mind as it'll take longer to download and have a larger footprint than the tarballs, while we can't actually enjoy the flexibility. But oh well.. That's pretty minor.

Versioning management, which was already bad, sadly becomes a pita. We'd want a way to centralize versioning to have it propagate. I'm starting to want to destroy everything to accommodate your changes better.. I'll think about it and do some refactoring to make things smoother for the both of us.

AdelKS:
I may have an idea: we can start tagging kernel subversions here in git. So we if we want to build 5.7.4, we git clone on that tag then use the cloned folder as we do now.

AdelKS:
I updated my previous message. I have the bad habit of writing up half baked ideas then edit my message with more thought out ideas.

AdelKS:
Oh maybe an even better idea: using git to version the patches, so `install.sh` and `PKGBUILD` would do:

```shell
git clone -b v5.7.8 https://patch-repo linux57-tkg-patches
```

Problem solved.

Tk-Glitch:
You're asking for a complete switch to something else. You should probably go for a fork - or an all new project - at this point :D

I'm making patches and kernel installers for Arch, only tested on Arch, using Arch's config. Having a script to package for other distros is one thing, this is something else entirely, with different paths I will never want to maintain.

What you want is the patches, so take the patches. There's nothing else being of interest here for what you want to achieve. I'll gladly add a link to your repo if you want me to :)

AdelKS:
> You're asking for a complete switch to something else. You should probably go for a fork - or an all new project - at this point :D

Haha in my first message I went full retard on wild ideas, though from my point of view I was in the brainstorming phase, you gave too much credit to the ideas I shot at you, my point was to give you even more ideas to consider. I didn't expect you to say "YEA LET'S GO!". Though actually if you did I would've probably tagged along xD

> I'm making patches and kernel installers for Arch, only tested on Arch, using Arch's config. Having a script to package for other distros is one thing, this is something else entirely, with different paths I will never want to maintain.

Absolutely, in the end the day the only extra thing the multi-branch idea would've enabled was being able to chose kernel subsubversions using git. Too huge of a change for something so small. So, agreed, forget about it :D

> What you want is the patches, so take the patches. There's nothing else being of interest here for what you want to achieve. I'll gladly add a link to your repo if you want me to :)

No no, I will tag along with you guys. End of story. I think dividing the work force too much is what was and is preventing the FOSS community to progress way faster than it could've. Plus, this is your work, I think it should stay here. And finally, I think your work is worth expanding to other Linux distros, to benefit a bigger user base.

So, to summarize:
- I want to make your work more usable out-of-the-box for other distros. Using my own time. Then I am planning to tag along to help you maintain the repo (at least on the script I provided, of course the decision would be yours).
- You want to spend the least amount of work on maintaining this repo. And so do I, I have a big project of mine to work on anyway :P

So, we can already stop at patching and configuring the latest Linux kernel. For compiling and installing, people can look that up for their own distros. This part is already done in my script.

One problem is left: the latest Linux version doesn't always work, so people will need to get your previous patches to be able to make a working kernel. That would require digging deeper into the inner workings of this repo. I had another suggestion in my other message but it is actually too big of a change too. 
How about adding subfolders in `linux57-tkg-patches` to keep older patches, one having a name that translates to `latest` that is used in PKGBUILD, the others would be there for easier old-patch lookup. I think this fits with you doing the least amount of work. What do you think ?

SveSop:
> With the latest sub-versions of linux-5.7, 5.7.6 and 5.7.7 so far (5.7.5 works) I have been getting these errors, they happen when the `.deb` files are being created, I can't find anything about it on the webs:
> 
> > error: include/uapi/linux/vt.h: leak CONFIG_NR_TTY_DEVICES to user-space
> > make[6]: *** [scripts/Makefile.headersinst:63: usr/include/linux/vt.h] Error 1
> > make[6]: *** Waiting for unfinished jobs....
> > HDRINST usr/include/linux/kvm_para.h
> > make[5]: *** [Makefile:1227: headers] Error 2
> > make[4]: *** [scripts/Makefile.package:87: intdeb-pkg] Error 2
> > make[3]: *** [Makefile:1463: intdeb-pkg] Error 2
> > make[2]: *** [debian/rules:9: binary-arch] Error 2
> > dpkg-buildpackage: error: fakeroot -u debian/rules binary subprocess returned exit status 2
> > make[1]: *** [scripts/Makefile.package:77: deb-pkg] Error 2
> > make: *** [Makefile:1463: deb-pkg] Error 2
> 
> Any idea what the reason behind this can be ?

The same type of error occurs when using the PDS patchset (i have not tried MuQSS or BMQ).
ie. `error: include/uapi/linux/sched.h: leak CONFIG_SCHED_PDS to user-space`

I fixed this by adding sched.h to a "ignore list" that bindeb-pkg uses to generate packages. I would assume that this is running some distro specific tests or whatnot.

Created this on my local fork: [https://github.com/SveSop/linux-tkg/commit/44841a0ffdef28f532eeff827911e2a035b5b841](https://github.com/SveSop/linux-tkg/commit/44841a0ffdef28f532eeff827911e2a035b5b841)
Not sure if it is the right spot to put it, since it seems (as the error above) Ubuntu only... 

I would also think you could do the same for vt.h and add
`include/uapi/linux/vt.h:CONFIG_NR_TTY_DEVICES`

This is not the correct way to fix it i would say, but then again, i dont really know why this happens, and asking packaging questions on ubuntu/debian forums on this would probably not fly at all :)

Feel free to test this "hack" for CONFIG_NR_TTY_DEVICES tho @AdelKS

AdelKS:
Hello @SveSop, thank you for this precious information! I see that you are also maintaining a repo for building a custom kernel for Ubuntu (I haven't found it while researching all this) for nearly two years. Keep up the good work! Don't forget to share your rare patch finds here :D

It seems that Tk-Glitch did something that actually will fix this problem. Because as I understand things now, after reading your message, is that Ubuntu's packaging script does some checks to see if some defaults have been changed, like `CONFIG_NR_TTY_DEVICES`. So if we don't touch them, we good. If Tk-Glitch's fixes don't work, I will dig deeper into your solution. Maybe there is something in Debian's or Ubuntu's docs.

@Tk-Glitch  comon tell me that you did something to fix the issue xD I wouldn't have known otherwise if @SveSop haven't written here. I just happened to see the small mention of your commit above his message.

Tk-Glitch:
> tell me that you did something to fix the issue

I have added a workaround for the specific patch you reported as problematic, so you could disable it easily with a switch. But if it can be fixed broadly I might as well revert that change :smiley:

SveSop:
I believe this is what is causing the debian package creation to fail rather than warn...
[https://github.com/torvalds/linux/commit/5967577231f9b19acd5a59485e9075964065bbe3](https://github.com/torvalds/linux/commit/5967577231f9b19acd5a59485e9075964065bbe3)

I am not entirely sure how other .deb maintainers do this, cos i just tried to compile kernel with BMQ and it failed aswell... So there is probably some switch someplace that can easily fix this :) I mean, you got Xanmod... Liquorix.. providing .deb pre-built kernels, so if this was a "hard failure" it would have been mentioned i guess :)

I will dig a wee bit more into this. The reason i have not caught this is that i mostly have been working on tweaking the default CFS scheduler a bit, and has not really bothered to test that many other schedulers.

AdelKS:
I am actually currently looking into it too!

I tried to find documentation on Debian or Ubuntu but didn't find anything. My current guess is that only the additional config options from the patches are marked as leaks, if they didn't exist in the linux tree to begin with. Or something close to that idea.

> I mean, you got Xanmod... Liquorix.. providing .deb pre-built kernels, so if this was a "hard failure" it would have been mentioned i guess :)

Maybe they are using the "ignore" trick in `scripts\headers_install.sh`

In the other hand, for example graysky's patches that adds various `-march=` optimizations introduce additional config options but they are not marked as leaks. Maybe that will help us understand how to make them work out.

> I believe this is what is causing the debian package creation to fail rather than warn...
torvalds/linux@5967577

They say "Misuse of CONFIG_* in UAPI headers should result in an error", so I suppose something is not done right in the patches. Any knowledge on that @Tk-Glitch ?

SveSop:
> I am actually currently looking into it too!
> 
> I tried to find documentation on Debian or Ubuntu but didn't find anything. My current guess is that only the additional config options from the patches are marked as leaks, if they didn't exist in the linux tree to begin with. Or something close to that idea.
> 
> > I mean, you got Xanmod... Liquorix.. providing .deb pre-built kernels, so if this was a "hard failure" it would have been mentioned i guess :)
> 
> Maybe they are using the "ignore" trick in `scripts\headers_install.sh`
> 
> In the other hand, for example graysky's patches that adds various `-march=` optimizations introduce additional config options but they are not marked as leaks. Maybe that will help us understand how to make them work out.
> 
> > I believe this is what is causing the debian package creation to fail rather than warn...
> > [torvalds/linux@5967577](https://github.com/torvalds/linux/commit/5967577)
> 
> They say "Misuse of CONFIG_* in UAPI headers should result in an error", so I suppose something is not done right in the patches. Any knowledge on that @Tk-Glitch ?

It does seem as Liquorix does this tweak to the headers_install.sh script.
[https://github.com/damentz/liquorix-package/blob/5.7/linux-liquorix/debian/patches/zen/v5.7.9-lqx1.patch](https://github.com/damentz/liquorix-package/blob/5.7/linux-liquorix/debian/patches/zen/v5.7.9-lqx1.patch)

```
@@ -94,22 +94,24 @@ include/uapi/linux/eventpoll.h:CONFIG_PM_SLEEP
 include/uapi/linux/hw_breakpoint.h:CONFIG_HAVE_MIXED_BREAKPOINTS_REGS
 include/uapi/linux/pktcdvd.h:CONFIG_CDROM_PKTCDVD_WCACHE
 include/uapi/linux/raw.h:CONFIG_MAX_RAW_DEVS
+include/uapi/linux/sched.h:CONFIG_SCHED_MUQSS
 "
```
So, they add CONFIG_SCHED_MUQSS in their patchset for Liquorix kernel.

AdelKS:
Well, it is the easy way out as of now. I mean if it doesn't do any harm otherwise (by that I mean we don't create silent errors or something when using the custom kernel), we have no reason to not use the ignore in `headers_install.sh`.

SveSop:
Well... there is probably some minor thing that does this tbh, cos if you take a look at the header file sched.h
`cat /usr/src/linux-headers-5.7xxxxxxxx/include/linux/sched.h | grep -i config_`
you will see that there is a lot of ifdef's and whatnot containing various config_ statements. Interesting thing tho.. Maybe Alfred Chen (BMQ) would know what this could be?

I'll see if i can make a question to him and see what his take on this would be :)

AdelKS:
> EDIT: This error does not seem to happen with BMQ/ProjectC tho... Hmm

That's one more pointer that says that tkg's patches need some kind of patch with respect to that. What are exactly the patch names from BMQ that work well ? `0009-prjc_v5.7-r2.patch` ?

SveSop:
@AdelKS 
I have not tested all the TkG patches together, so i cannot guarantee what works or not.. I kinda pulled the prjc patches from Alfred's gitlab..

Anyway, i do think i found the reason this happens.

1. The upstream kernel patch i mentioned earlier was from may. [https://lkml.org/lkml/2020/5/5/1480](https://lkml.org/lkml/2020/5/5/1480)
This means that it made a "stricter" check of this i guess, and instead of just a warning, it generated an error - and thus stopping the debian buildscript for .deb packages.

2. I was kinda looking at the wrong spot - I am no programmer at all, so kinda winging this - The "problem" this checks is that the PDS uapi contains this change to `include/uapi/linux/sched.h`
```
@@ -115,7 +115,10 @@ struct clone_args {
 #define SCHED_FIFO		1
 #define SCHED_RR		2
 #define SCHED_BATCH		3
-/* SCHED_ISO: reserved but not implemented yet */
+/* SCHED_ISO: Implemented in BFS/MuQSSPDS only */
+#ifdef CONFIG_SCHED_PDS
+#define SCHED_ISO		4
+#endif
 #define SCHED_IDLE		5
 #define SCHED_DEADLINE		6
```
Aside from the minor spelling mistake, skipping a / there, there should be no real need to do a check for this CONFIG_ here.. I would think the only "harm" would come from defining something that would possibly not be used would be a compile warning about this. I do not think it would break CFS if SCHED_ISO would be defined, but i do not know. I will atleast test a compile about this with default CFS WITH this defined..

So, simply just removing the ifdef lines like this, fixes it:
```
@@ -115,7 +115,10 @@ struct clone_args {
 #define SCHED_FIFO		1
 #define SCHED_RR		2
 #define SCHED_BATCH		3
-/* SCHED_ISO: reserved but not implemented yet */
+/* SCHED_ISO: Implemented in BFS/MuQSS/PDS only */
+#define SCHED_ISO		4
 #define SCHED_IDLE		5
 #define SCHED_DEADLINE		6
```
Probably need a bit more educated guesswork than what i do....

AdelKS:
Thank your for this precious information again!

I patched the linux kernel with BMQ/ProjectC in one side and PDS-undead in the other and I am running diffs, I am supposing BMQ/ProjectC patch defines everything properly so there's no "leak" whatsoever as you told me. The additional information you provided will for sure help me figure this out. Because other leaks are happening too when using the other patches. I will keep you updated. I will keep working on it later tonight or tomorrow.

SveSop:
I have posted a question on -CK's blog asking for thoughts about this.
[https://ck-hack.blogspot.com/2020/06/linux-57-ck1-muqss-version-0202-for.html?showComment=1595085837200#c4079758906353271215](https://ck-hack.blogspot.com/2020/06/linux-57-ck1-muqss-version-0202-for.html?showComment=1595085837200#c4079758906353271215)

I did a compile of a regular CFS scheduler kernel WITH define SCHED_ISO 4, and it gave no warning, but what would happen if you actually tried to schedule a task as "ISO", i do not know tbh. For now (for my own kernel patch-set) i will implement defining this on PDS no matter, cos i would think/hope whoever bothers using the patchset actually did it to run PDS.

I mean, there is no reason to patch the kernel with a scheduler and NOT use it - and from what i gather of the script setup here, you opt-in if you want to use a particular scheduler.. no?

AdelKS:
I will implement your trick in tkg undead-pds patch, best solution so far since as you said, it's an opt-in patch that, when chosen, does activate PDS scheduler automatically.

>I did a compile of a regular CFS scheduler kernel WITH define SCHED_ISO 4, and it gave no warning, but what would happen if you actually tried to schedule a task as "ISO"

So this problem does not happen in our case, I will try it out.

AdelKS:
Works like a charm ! I can finally continue my work :D

AdelKS:
Hello Mister @Tk-Glitch,

I consider my work for Debian derivatives as done, for Fedora I'd like to do another PR so this is already available and I'll take more time for Fedora, though it won't need as much work. I think it is time I take a bit of your time for reviewing!

For the git shallow cloning instead of unpacking a tar file, I think for non-arch distros it's better, given my current knowledge though: at each script run I need to reset the sources to their initial state, I don't know how to do that with tars except unpack them again on a clean folder and delete the old one (overwrite amounts to the same no ? with extra files staying there though), which takes out more SSD disk life more than a `git clean` + `git reset --hard`, and is slower. What do you think ?

Adel

Tk-Glitch:
I'll review soon.

linux57-tkg/README.md
```diff
@@ -32,8 +32,38 @@ You can enable support for it at the beginning of the PKGBUILD file. Make sure t
...
 - **optional** Fsync support (proton)
 
+
+## Install procedure
+
+### Arch & derivatives
 ```
 git clone https://github.com/Frogging-Family/linux-tkg.git
 cd linux-tkg/linux57-tkg
+# Edit customization.cfg file
```
Tk-Glitch, line 42: Is the header formatting intentional here? It feels out of place.

linux57-tkg/customization.cfg
```diff
@@ -77,9 +79,6 @@ _ftracedisable="false"
 # Set to "true" to disable NUMA, lowering overhead, but breaking CUDA/NvEnc on Nvidia equipped systems - Kernel default is "false"
 _numadisable="false"
 
-# Set to "true" to enable misc additions - May contain temporary fixes pending upstream or changes that can break on non-Arch - Kernel default is "true"
-_misc_adds="true"
```
Tk-Glitch, line 81: This is here to stay, even if I end up removing the specific patch it was initially introduced for.

linux57-tkg/linux57-tkg-config/prepare
```diff
@@ -1,68 +1,229 @@
...
-    patch -Np1 -i ../0009-glitched-bmq.patch
+
+    msg2 "Applying Glitched BMQ patch"
+    patch -Np1 -i "$_where"/0009-glitched-bmq.patch
+    msg2 "Done"
   else
-    patch -Np1 -i ../0003-glitched-cfs.patch
+    msg2 "Applying Glitched CFS patch"
+    patch -Np1 -i "$_where"/0003-glitched-cfs.patch
+    msg2 "Done"
```
Tk-Glitch, line 216: I'm not convinced by those "Done" messages. Am I missing something?

linux57-tkg/linux57-tkg-config/prepare
```diff
@@ -1,68 +1,229 @@
...
-  patch -p1 -i ../patch-"${pkgver}"
-
-  # ARCH Patches
-  if [ "${_configfile}" == "config_hardened.x86_64" ] && [ "${_cpusched}" == "cfs" ]; then
-    msg2 "Using linux hardened patchset"
-    patch -Np1 -i ../0012-linux-hardened.patch
+_define_vars() {
+ 
+  _basever=57
+  _basever_date="31/05/2020"
```
Tk-Glitch, line 6: I will never update this. Is this actually required for anything?

linux57-tkg/linux57-tkg-patches/0005-v5.7_undead-pds099o.patch
```diff
@@ -569,9 +569,9 @@ index 3bac0a8ceab2..d6d384ddb57d 100644
  #define SCHED_BATCH		3
 -/* SCHED_ISO: reserved but not implemented yet */
 +/* SCHED_ISO: Implemented in BFS/MuQSSPDS only */
-+#ifdef CONFIG_SCHED_PDS
```
Tk-Glitch, line 572: If it's okay for you I'd prefer if those changes were part of another PR.

linux57-tkg/README.md
AdelKS: I am not sure I understood, you mean the space after the "#" ?

linux57-tkg/customization.cfg
AdelKS: Yea I wanted to discuss that with you. As I understand now you didn't put it only for Ubuntu to be able to compile, then okay! 

Are you the one who reverts commits or me ?

linux57-tkg/linux57-tkg-config/prepare
AdelKS: I was debugging in the beginning, and I wanted to know which lines are terminating properly. If you don't like the "Done" let's get rid of it!

linux57-tkg/linux57-tkg-config/prepare
AdelKS: I just forgot them them there x) They are not used anywhere.

linux57-tkg/README.md
Tk-Glitch: Oh I see, you wanted to explicitly comment it out. Okay.

linux57-tkg/linux57-tkg-patches/0005-v5.7_undead-pds099o.patch
AdelKS: Okay no problem! This is another fix so Ubuntu can compile.

linux57-tkg/customization.cfg
Tk-Glitch: Yeah it was to have a place to add temporary patches without using -base for it. The decoupling of "that patch" with -base was following that logic of temporary fixing the issue we had with it.

I have no power on the PR's code until it's effectively merged afaik.

linux57-tkg/linux57-tkg-config/prepare
Tk-Glitch: I haven't actually ran the script to see the output yet, but with the patch coming after outputting it's "header message" I gave you my initial thoughts about it. I'm not against them if you consider they are helping out.

linux57-tkg/linux57-tkg-config/prepare
Tk-Glitch: Good :+1:

linux57-tkg/linux57-tkg-patches/0005-v5.7_undead-pds099o.patch
Tk-Glitch: I followed that despite my silence :stuck_out_tongue: But since it's not directly related to the ubuntu building implementation and more of a misc source fix, I feel like it should be separate for proper indexing.

AdelKS: I guessed so! Since you were still active in this repository :sunglasses:

linux57-tkg/customization.cfg
AdelKS: I actually gave you push permissions hehe

linux57-tkg/linux57-tkg-patches/0005-v5.7_undead-pds099o.patch
AdelKS: Done!

linux57-tkg/linux57-tkg-config/prepare
AdelKS: Fixed.

linux57-tkg/linux57-tkg-config/prepare
AdelKS: I find them useful and like to see that something is "Done" before it moves to something else. The final decision is up to you, it will take me few seconds to remove all the "Done" messages.

linux57-tkg/customization.cfg
AdelKS: I have a small concern about the misc patch, I would like people from ubuntu for example to get the amdgpu patch but not have the change that gives the leaky tty :/

AdelKS:
Now that Linux 5.8 is out, I will add the install.sh there too.

Tk-Glitch:
I'm merging the 5.7 version and will work on making it a bit more flexible.

AdelKS:
Awesome! Will you make the script for 5.8 or do you want me to do it ?

Tk-Glitch:
I'll reflect the changes to 5.8, no worries. The small tweaks I have made in the commit above should be all good on buntu but I haven't tested them there, so I might have missed a quirk or two. Do you happen to have checked that I didn't break anything by any chance ? :smile:

AdelKS:
Cool! I haven't tested the Ubuntu build with your changes, will do now. Also can you remove all occurrences of `_command_recognised` in `install.sh` ? Forgot it there after I changed the way I handled non-recognized commands.

AdelKS:
Sorry, another thing, in `customization.cfg`, `CUSTOM_GCC_PATH` is not Arch specific anymore. `Install.sh` does use it.

AdelKS:
Okay it works! I got bamboozled by the `_misc_adds` variable: I waited for the entirety of the compile process to get the leaky message error. I think it is a good idea to make an interactive prompt for it like the other options :smile:

PR #31, opened 2020-07-14
linux57-tkg: Import hardened config and patchset

bcvandendool:
Config: https://www.archlinux.org/packages/extra/x86_64/linux-hardened/ 
Patchset: https://github.com/anthraxx/linux-hardened

The hardened patchset has recently been released for 5.7. I am currently running it, and seems to have no problems so far.


Tk-Glitch:
Thank you for the heads up! I'll include it soon.

PR #36, opened 2020-07-30
linux57: Fix leak error for CONFIG_PDS in make deb-pkg

AdelKS:
Here's the PR for these commits alone, I think they won't introduce any regression in this repository because the PDS patch is only used when PDS is chosen, which automatically does the define.


Tk-Glitch:
Thank you!

AdelKS:
Happy yo help :D

PR #37, opened 2020-08-04
linux58-tkg: Update from RC7 to release

FLGX06:
I hope this is everything that is needed for update from RC to release except for the config file.
I didn't touch the config file because I'm not sure how to do that one. (I'm getting `Support for P6_NOPs on Intel chips (X86_P6_NOP) [N/y/?] (NEW)` question I have to set)
I also removed 4 amdgpu hunks from misc_additions.patch, which it says were already applied (I assume the patch got accepted upstream)
Hope this makes it easier for you to update it to release even though it's simple PR! :frog:


Tk-Glitch:
This is looking pretty good. I was planning to do it today. Thanks for helping out :)

PR #41, opened 2020-08-08
linux57: extra changes to Ubuntu install script that have been forgotten

AdelKS:
Hello!

I forgot to do some changes, so here they are :D


Tk-Glitch:
:+1: Thanks

PR #42, opened 2020-08-08
linux57: add interactive prompt for _misc_adds

AdelKS:
I think this is a quality of life improvement for the end user, especially Ubuntu guys (maybe Fedora, I started working on it, I will see if it gets those leak errors), since they will for sure go through the compile process once with `_misc_adds=true`, see that it doesn't work. Try to find out why, and then maybe end up trying `_misc_adds=false`. 

I believe having the prompt would help. What do you think ?

Edit: Fedora gets the leaky TTY too, maybe I should work on a fix for the leak, like PDS actually


Tk-Glitch:
I have nothing against this, thanks!

AdelKS:
Happy to help!

PR #47, opened 2020-08-19
Accept default options to "make localmodconfig"

CodingKoopa:
This PR pipes `yes ""` to `make localmodconfig`, to alleviate Tk-Glitch/PKGBUILDS#404. I deem this to be an acceptable solution as this seems to be "not our bug" as this affects more than TkG, and the options that are prompted for seem to be pretty insignificant.


Tk-Glitch:
I consider this as an acceptable workaround as well. It doesn't fix the other issues modprobed-db has, of course :D
Thank you!

PR #49, opened 2020-08-19
Install the TkG configuration used to the system

CodingKoopa:
This pull request adds basic functionality for saving TkG variables (variables preceded with a `_`) to a file, and installing that file to the system. This can be useful for testing different kernel configurations, and seeing what the current version was compiled with. I have only implemented this for 5.7 as there are some design concerns that are up in the air:
- As is, `hackheaders` removes documentation from the package, so the only "documentation" produced by linux-tkg in this PR is `/usr/share/doc/${pkgbase}/customization.cfg`. As such, if one has `options=(!docs)` in their `makepkg.conf`, here it would only serve to strip that `customization.cfg`, so I have overrode that.
  - Is it desired behavior to produce this file if even if someone has `options=(!docs)`?
  - Would this file belong in `/usr/share/doc/` at all?
- Should this functionality be put behind a configuration variable?
  - Does this functionality belong here at all?
- This includes the path in which the package was compiled, which is potentially sensitive information, considering people can share packages.
- Currently, this includes some variables which are "internal" to the script. Is this acceptable?

These aren't questions I all expect answers to, I'm just laying out my thoughts as this is currently an RFC PR.

[Here](https://pastebin.com/66Ms7t7e) is an example of the file installed to the system.

### DIFF
--- a/linux54-tkg/PKGBUILD
+++ b/linux54-tkg/PKGBUILD
@@ -105,7 +105,7 @@ url="http://www.kernel.org/"
 license=('GPL2')
 makedepends=('xmlto' 'docbook-xsl' 'kmod' 'inetutils' 'bc' 'libelf' 'patchutils' 'flex' 'python-sphinx' 'python-sphinx_rtd_theme' 'graphviz' 'imagemagick' 'git')
 optdepends=('schedtool')
-options=('!strip')
+options=('!strip' 'docs')
 source=("https://www.kernel.org/pub/linux/kernel/v5.x/linux-${_basekernel}.tar.xz"
         "https://www.kernel.org/pub/linux/kernel/v5.x/patch-${pkgver}.xz"
         "https://raw.githubusercontent.com/graysky2/kernel_gcc_patch/master/enable_additional_cpu_optimizations_for_gcc_v10.1%2B_kernel_v4.19-v5.4.patch"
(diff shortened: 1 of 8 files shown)


PedroHLC:
Hey @Tk-Glitch
I like this idea a lot!

Tk-Glitch:
We're doing something similar for wine/proton-tkg and while there isn't as many possible config permutations, I still consider this as being useful.

>  Is it desired behavior to produce this file if even if someone has options=(!docs)?

I would say it is.

>  Would this file belong in /usr/share/doc/ at all?

I personally don't see an issue with it being in the lib/modules/$kernel dir directly, but for cleanliness, /usr/share/doc/ works IMO.

> Should this functionality be put behind a configuration variable?

I don't think it's necessary, but I have a small issue with it I'll talk about below.

> Does this functionality belong here at all?

Sure.

> This includes the path in which the package was compiled, which is potentially sensitive information, considering people can share packages.
> Currently, this includes some variables which are "internal" to the script. Is this acceptable?

That's the issue I have with that currently, that I mentioned above.
We'd want both `_EXT_CONFIG_PATH` and `_where` values to be stripped out.

That's about it from me :+1:

CodingKoopa:
Updated for all current linux-tkg versions, and added a note to the readmes.

Tk-Glitch:
All good for me :+1: 
Thank you! :frog:

PR #55, opened 2020-08-23
Fix certificate issue on Debian

iWeaker4you:
I only applied it in version 5.8 so that first you can see if it applies to all, or if my solution is viable.

The problem is that every time it does the verification, it downloads the original configuration from github and does not maintain its own manual configuration that solves the certificate problem in Debian, it does not finish compiling for that reason.

And when you give Install.sh the CONFIG_SYSTEM_TRUSTED_KEYS = "debian / certs / test-signing-certs.pem" is reset, regardless of whether you edited it.

### DIFF
--- a/linux58-tkg/install.sh
+++ b/linux58-tkg/install.sh
@@ -46,7 +46,7 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
 
   source linux*-tkg-config/prepare
 
-  if [ $1 = "install" ] && [[ "$_distro" != "Ubuntu" || "$_distro" != "Debian" ]]; then
+  if [ $1 = "install" ] && [[ "$_distro" != "Ubuntu" && "$_distro" != "Debian" ]]; then
     msg2 "Variable \"_distro\" in \"customization.cfg\" hasn't been set to \"Debian\" or \"Ubuntu\""
     msg2 "This script can only install custom kernels for Ubuntu and Debian derivatives. Exiting..."
     exit 0
--- a/linux58-tkg/install.sh
+++ b/linux58-tkg/install.sh
@@ -46,13 +46,13 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
 
   source linux*-tkg-config/prepare
 
-  if [ $1 = "install" ] && [ "$_distro" != "Ubuntu" ]; then
-    msg2 "Variable \"_distro\" in \"customization.cfg\" hasn't been set to \"Ubuntu\""
+  if [ $1 = "install" ] && [ "$_distro" != "Ubuntu" ] && [ "$_distro" != "Debian" ]; then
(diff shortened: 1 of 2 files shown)


AdelKS:
Hello,

> The problem is that every time it does the verification, it downloads the original configuration from github and does not maintain its own manual configuration that solves the certificate problem in Debian, it does not finish compiling for that reason.

It does not download the `.config` file from Github but copies the one that is in `/boot/config-`uname -r`.config`: It does NOT use `linux${_basekernel//.}-tkg-config/config.x86_64`. So the commit as is won't fix anything and the `.config` file will get overwritten later in the script by this code chunk: 

```shell
msg2 "Copying current kernel's config and running make oldconfig..."
cp /boot/config-`uname -r` .config
yes '' | make oldconfig
msg2 "Done"
```
After you correct your previous commit, I will have to try if it doesn't break on Ubuntu. If it does I think we'd need to add a new option for `_distro` in `customization.cfg`, which would be `debian`. Also, which Debian are you using ? Will it work for the other versions ?

iWeaker4you:
@AdelKS  I have Debian 11 Testing, I tested it a while ago in 10, and I still have the certificate problem, I still think that it is best to put an extra option for Debian, because in Ubuntu it does not happen, in Ubuntu I tested it 3 days ago and it works without that problem, let me correct it.

iWeaker4you:
@AdelKS  Do you have Ubuntu?
Can you check if it works?

AdelKS:
Your change as-is will also affect Ubuntu, one needs to introduce Debian and do all the changes needed. @Tk-Glitch needs to give his thoughts on this too.

iWeaker4you:
@AdelKS  can you try again bro?

linux58-tkg/install.sh
```diff
@@ -46,13 +46,13 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
 
   source linux*-tkg-config/prepare
 
-  if [ $1 = "install" ] && [ "$_distro" != "Ubuntu" ]; then
-    msg2 "Variable \"_distro\" in \"customization.cfg\" hasn't been set to \"Ubuntu\""
+  if [ $1 = "install" ] && [ "$_distro" != "Ubuntu" ] && [ "$_distro" != "Debian" ]; then
```
AdelKS, line 49: Here you need an `or` too: It needs to be either Ubuntu OR Debian

AdelKS (commented):
I think this is the last needed change :thinking: then @Tk-Glitch will have his final say :D

linux58-tkg/install.sh
AdelKS: My bad, I went too fast and haven't read what came after the if condition... It was good. Sorry about that.

Tk-Glitch:
Looks fine to me :+1:

PR #56, opened 2020-08-25
Support Fedora and derivatives

AdelKS:
Hello!

I am done with the Fedora script, I although have an issue: I only tried the script in a VM with Gnome Boxes, and after installing with `dnf`, the new kernel doesn't show on the boot entries, and worse: if I try to uninstall it, I wants to remove all the kernel dependencies... I think this is specific to my VM but I don't know, it needs testers :D


linux57-tkg/install.sh
```diff
@@ -137,7 +142,14 @@ if [ "$1" = "install" ]; then
       export CCACHE_SLOPPINESS="file_macro,locale,time_macros"
       export CCACHE_NOHASHDIR="true"
       msg2 'ccache was found and will be used'
+
+    elif [ "$_distro" = "Fedora" ] && dnf list --installed ccache > /dev/null; then
+      export PATH="/usr/lib64/ccache/:$PATH"
+      export CCACHE_SLOPPINESS="file_macro,locale,time_macros"
+      export CCACHE_NOHASHDIR="true"
```
CodingKoopa, line 149: Since this looks to be the same between distributions, this could be moved into the outer `if`.

linux57-tkg/install.sh
```diff
@@ -137,7 +142,14 @@ if [ "$1" = "install" ]; then
...
+      msg2 'ccache was found and will be used'
     fi
+
```
CodingKoopa, line 152: Unneeded newline?

linux57-tkg/install.sh
```diff
@@ -164,6 +176,36 @@ if [ "$1" = "install" ]; then
...
+        msg2 "Install successful"
+
+        # Add to the list of installed kernels, used for uninstall
+        if ! { [ -f installed-kernels ] && grep -Fxq "$_kernelname" installed-kernels; }; then
+          cd "$_where"
+          echo $_kernelname >> installed-kernels 
+        fi   
+      fi
+    fi
+  fi
```
CodingKoopa, line 208: Perhaps make this an `elif` with the Ubuntu installation, since they are analogous?

linux57-tkg/install.sh
```diff
@@ -164,6 +176,36 @@ if [ "$1" = "install" ]; then
...
+
+  # Replace dashes with underscores, it seems that it's being done by binrpm-pkg
+  # Se we can actually refer properly to the rpm files.
+  if [ "$_distro" = "Fedora" ]; then
+    _kernel_flavor=${_kernel_flavor//-/_}
+  fi
+
+  if [ "$_distro" = "Fedora" ]; then
+    if make -j ${_thread_num} binrpm-pkg EXTRAVERSION="_${_kernel_flavor}"; then
+    #if [ "0" = "0" ]; then
```
CodingKoopa, line 188: Leftover debug code?

linux57-tkg/install.sh
AdelKS: Agreed. Will do.

linux57-tkg/install.sh
AdelKS: Yes xD

linux57-tkg/install.sh
AdelKS: I'll nuke™ it

linux57-tkg/install.sh
AdelKS: Will do, needs a small tweak, you tell me what you think of it. I will push the change in a few minutes

CodingKoopa:
Thanks, code looks fine to me now.

kostadin04:
It compiled successfully for me with the PDS scheduler, but when I try MuQSS, it gives me an error:

error: include/uapi/linux/sched.h: leak CONFIG_SCHED_MUQSS to user-space
make[3]: *** [scripts/Makefile.headersinst:63: usr/include/linux/sched.h] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:1230: headers] Error 2
make[1]: *** [scripts/Makefile.package:66: binrpm-pkg] Error 2
make: *** [Makefile:1466: binrpm-pkg] Error 2

AdelKS:
Hello @kostadin04,

Thanks for the feedback! It seems that MuQSS has a known issue we're trying to fix, I'll see if I can fix that in a different PR.

Also, did the PDS kernel install and boot properly? How about the uninstall?

kostadin04:
The PDS kernel didn't install at first, because i was missing the `rsync` package, but after installing it, the script finished with no problems at all, it booted fine and there were no issues present. When trying to uninstall it, the kernel was removed, but it removed a lot of packages other than the kernel. 
[List of uninstalled packages.txt](https://github.com/Frogging-Family/linux-tkg/files/5129762/List.of.uninstalled.packages.txt)

AdelKS:
Thanks for the feedback! I will add `rsync` to the dependency installs. About the packages that it wants to uninstall, I am pretty sure it should not. Did the original kernel get overwritten ? I suppose you are running a VM install too ? Because if the original kernel is overwritten then it's kinda "normal" that `dnf` is trying to uninstall those packages since no other packages needs them anymore.

kostadin04:
I am not running it in a VM, and the original kernel is still present, it hasn't been overwritten.

AdelKS:
Noted! That's good news that the original didn't get overwritten :D Because that would've been bad. 

About the packages that get uninstalled, I don't understand why they are getting uninstalled, some of them have been "manually" installed by the script, which means they should be considered as "user installed" by `dnf` and not remove them even if they are not needed by any other package ? I may be wrong :thinking: what I can do is make `dnf` only uninstall the kernel and its headers without removing anything else.

kostadin04:
I was able to remove them by doing `rpm -qa | grep "kernel-headers"
` and `rpm -qa | grep "kernel"` and then uninstalling the TKG packages that popped up by doing `sudo rpm -e --nodeps (package name)`. I don't know if that's a good way of doing it. :grinning:

AdelKS:
I will go for the extra option `--noautoremove` with `dnf remove [...]`, it should work I think.

AdelKS:
The good thing with using `rpm` is that I believe it is available in all RPM based distros, so the support would be broader if I only use `rpm`, though dependency installation should still use each distro's package manager. So I don't know what do to, I learned that Opensuse uses Zypper, I am planning to use it and hope that package names are the same.

kerOssinas:
Hey @AdelKS !

Tested this on a Fedora 32 VM, install seems to work fine, had no problems but the uninstall with dnf doesn't work since kernel is a protected package, removed it with rpm, seemed to work but something fucky happens, I don't get the grub boot menu anymore and grub2-mkconfig doesn't detect anything just like it happened to you. Needs more testing why that happens.

Another thing is I don't think removing kernel-headers is a good idea, when you install it it replaces the older package and glibc depends on kernel-headers.

Also, an install on OpenSUSE TW is still running with some tweaks, will report later how it goes.

bkdwt:
> Hey @AdelKS !
> 
> Tested this on a Fedora 32 VM, install seems to work fine, had no problems but the uninstall with dnf doesn't work since kernel is a protected package, removed it with rpm, seemed to work but something fucky happens, I don't get the grub boot menu anymore and grub2-mkconfig doesn't detect anything just like it happened to you. Needs more testing why that happens.
> 
> Another thing is I don't think removing kernel-headers is a good idea, when you install it it replaces the older package and glibc depends on kernel-headers.
> 
> Also, an install on OpenSUSE TW is still running with some tweaks, will report later how it goes.

I'm so excited to know the result, as I am an OpenSUSE Tumbleweed user.

kerOssinas:
On OpenSUSE TW I changed the dependency installation in the script to:
`sudo zypper install -y rpmdevtools ncurses-devel pesign libXi-devel gcc-c++ git ccache flex bison elfutils libelf-devel openssl-devel dwarves make patch bc rpm-build`

After it built it I installed only the kernel rpm manually with rpm -i and it seems to work.

bkdwt:
Are you using the 5.7 or 5.8 tkg kernel @ker0ssinas? Thanks for the tip.

kerOssinas:
@bkdwt The 5.7 since @AdelKS made the script for 5.7 but it should work for 5.8.

bkdwt:
Ok, I'll try the 5.7 version here. Thanks!

AdelKS:
Hello guys!

Thank you for all the precious information ! :D

> I don't get the grub boot menu anymore and grub2-mkconfig doesn't detect anything just like it happened to you. Needs more testing why that happens.

Okay it seems I will have to use an USB and install fedora on it to see if the VM is doing something fishy.

> Another thing is I don't think removing kernel-headers is a good idea, when you install it it replaces the older package and glibc depends on kernel-headers.

As far as I understand, the headers aren't supposed to replace the older ones ? Maybe I'm wrong. I do not remember Ubuntu's behavior regrading that.

> I'm so excited to know the result, as I am an OpenSUSE Tumbleweed user.

We will get it to work! What's sure is that I'm not gonna give up on it till it works flawlessly.

> On OpenSUSE TW I changed the dependency installation in the script to [...]

Thanks ! That's very useful :D

> Ok, I'll try the 5.7 version here. Thanks!

Once the script done for 5.7, I will mirror it for the next versions :)

I will be able to work on this tonight (meaning in around 9h) or tomorrow 😁

Once this is done, we definitely have to play a few games together with our new linux-tkg master race kernel :D

bkdwt:
> Hello guys!
> 
> Thank you for all the precious information ! :D
> 
> > I don't get the grub boot menu anymore and grub2-mkconfig doesn't detect anything just like it happened to you. Needs more testing why that happens.
> 
> Okay it seems I will have to use an USB and install fedora on it to see if the VM is doing something fishy.
> 
> > Another thing is I don't think removing kernel-headers is a good idea, when you install it it replaces the older package and glibc depends on kernel-headers.
> 
> As far as I understand, the headers aren't supposed to replace the older ones ? Maybe I'm wrong. I do not remember Ubuntu's behavior regrading that.
> 
> > I'm so excited to know the result, as I am an OpenSUSE Tumbleweed user.
> 
> We will get it to work! What's sure is that I'm not gonna give up on it till it works flawlessly.
> 
> > On OpenSUSE TW I changed the dependency installation in the script to [...]
> 
> Thanks ! That's very useful :D
> 
> > Ok, I'll try the 5.7 version here. Thanks!
> 
> Once the script done for 5.7, I will mirror it for the next versions :)
> 
> I will be able to work on this tonight (meaning in around 9h) or tomorrow grin
> 
> Once this is done, we definitely have to play a few games together with our new cross-distro custom linux-tkg kernel :D

You rock, dude!

bkdwt:
> On OpenSUSE TW I changed the dependency installation in the script to:
> `sudo zypper install -y rpmdevtools ncurses-devel pesign libXi-devel gcc-c++ git ccache flex bison elfutils libelf-devel openssl-devel dwarves make patch bc rpm-build`
> 
> After it built it I installed only the kernel rpm manually with rpm -i and it seems to work.

Can you share the install.sh file? Thanks!

kerOssinas:
@AdelKS I don't know how to reply to a comment directly on the phone but whatever.

Kernel-headers package replaces older kernel-headers package, kernel-devel though installs alongside other versions. When I build my own kernels I only installed kernel and kernel-devel rpms and it was enough to build Nvidia driver modules against kernel-devel which installs under /usr/src/.

kerOssinas:
@bkdwt I'm away this weekend, you can edit the install.sh yourself, where there is dependency installation replace the dnf commands with the one I provided and for the kernel installation change from dnf to rpm.

bkdwt:
@ker0ssinas just editing the dependencies doesn't work for me. The script asks for Ubuntu or Fedora, because the script doesn't have a Opensuse config.

AdelKS:
@bkdwt I'll do it tonight if you are not in a hurry :)

kerOssinas:
@bkdwt I didn't add an OpenSUSE option, I just made the Fedora version work for it, after the change just select Fedora.

I just did it to see if it would work, to make a legit OpenSUSE version you'd need a few more changes.

AdelKS:
> Kernel-headers package replaces older kernel-headers package

You know what happens if you run a kernel version but with headers installed meant for another kernel version?

kerOssinas:
@AdelKS I just know you can run newer headers with an older kernel since they're backwards compatible. I think since glibc depends on it some headers may be needed for compiling programs. I'm not an expert on this so not 100% sure.

AdelKS:
Okay, I'll make the script install/replace headers but not uninstall them, until a better solution is found, if there is any.

Kodehawa:
The kernel-headers don't seem to work:
Screenshot-20200829204509-841x196
Screenshot-20200829204523-684x310


Can't install rtw88 or rtl8821ce without proper headers, so kind of in a loss here. Thanks!

AdelKS:
Hello @Kodehawa,

Have you tried forcing the headers install with `rpm -i --replacepkgs`, even though they seem to be already installed.  I kinda have no clue why they're not found in `/lib/modules/5.7.19_tkg_pds+` but this seems related to the problems I am already having. currently working on it.

Kodehawa:
> Hello @Kodehawa,
> 
> Have you tried forcing the headers install with `rpm -i --replacepkgs`, even though they seem to be already installed. I kinda have no clue why they're not found in `/lib/modules/5.7.19_tkg_pds+` but this seems related to the problems I am already having. currently working on it.

No dice. Same error :(

AdelKS:
Can you check if you have anything in `/lib/modules/5.7.19_tkg_pds+` ? Also what does `uname -r` give you ? Are you in Fedora.

Kodehawa:
@AdelKS Yes, fedora. Yes, there's content on it, but there's no build directory on it, which is what Fedora wants.

AdelKS:
isn't dkms run when installing the kernel actually ? or wait, you need the `kernel-devel` package I think. Lemme look into that.

Kodehawa:
Yes. There's no kernel-devel package generated by this.

bkdwt:
Hey @AdelKS, tested your build on OpenSUSE Leap 15.2 but I get this error with automatically install:

```
-> Building successfully finished!
Do you want to install the new Kernel ? y/[n]: y
[sudo] password for root: 
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following 2 NEW packages are going to be installed:
  kernel kernel-headers

The following package requires a system reboot:
  kernel

2 new packages to install.
Overall download size: 69.2 MiB. Already cached: 0 B. After the operation, additional 309.2 MiB will be used.

    Note: System reboot required.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package kernel-headers-5.7.19_tkg_pds_lp152.36_default+-1.x86_64                          (1/2),   1.1 MiB (  5.0 MiB unpacked)
kernel-headers-5.7.19_tkg_pds_lp152.36_default+-1.x86_64.rpm:
    Package is not signed!

kernel-headers-5.7.19_tkg_pds_lp152.36_default+-1.x86_64 (Plain RPM files cache): Signature verification failed [6-File is unsigned]
Abort, retry, ignore? [a/r/i] (a): a
Problem occurred during or after installation or removal of packages:
Installation has been aborted as directed.
Please see the above error message for a hint.
 -> exit cleanup done
```

bkdwt:
Ok, installed manually and works. Now I waiting for linux58-tkg for a test with Tumbleweed. Thank you very much again for your work @AdelKS! :+1: 

Screenshot from 2020-08-30 03-57-57

AdelKS:
Hello @bkdwt,

Yes I had the same error, it's a simple error that can be avoided by telling `zypper` to not check for signature. I'm still working on it 😁

For tumbleweed, you can already use 5.7, or you specially need 5.8 ? In any case, the script for it will come!

Happy to help!

bkdwt:
@AdelKS It's because TW comes with I kernel 5.8 by default now, so I didn't want to use an older kernel. 😅

AdelKS:
Hello @bkdwt,

Got it! I think the Suse script is working now. I only have to add some extra things, same for Fedora and we should be good!

AdelKS:
@Kodehawa the kernel-devel package is installed by the script now, can you retry ?

bkdwt:
@AdelKS Compiling the 5.7 for try on TW on bare metal ;)

AdelKS:
Cool! Keep me updated @bkdwt :)

bkdwt:
> Cool! Keep me updated @bkdwt :)
rpmbuild is using only 1 core Oo

AdelKS:
> now the install works like a charm. Zero problems! :D
Happy to hear that ! I will soon make the script work for 5.8 :) then we can play games!

> rpmbuild is using only 1 core Oo
That step is very slow and I don't know how to change it 🤔

AdelKS:
@bkdwt, here you go, you can use the 5.8 script now :)

bkdwt:
> @bkdwt, here you go, you can use the 5.8 script now :)

Love ya (no homo!) hahahahaha 

I will compile 5.8 later. Don't forget to send your paypal ;)

AdelKS:
Happy to have helped :D I guess you don't happen to live in France so we can grab a drink, you can offer me a drink through [zegrapher's website](zegrapher.com) :P (there's a link at the bottom)

@Tk-Glitch, @CodingKoopa, you can start now the review. I think I am done. What's left is all the things I may have forgotten about, and you can remind me of them once you review. To summarize:
- Two new possibilities to the `_distro` variable: Fedora and Suse
  - The variable is not set by default now, and is prompted by the script.
- The uninstall part of the kernel has been removed, now it only shows the list of all installed kernels in the machine and hints to the user what to do to remove the kernels. Since it's a touchy subject, the user takes the blame if something goes wrong at uninstall xD
- The script across the kernel versions is now exactly the same, except for linux59 because the CPU opt patch still has 5.8 in its name.

linux57-tkg/install.sh
```diff
@@ -44,17 +47,46 @@ _misc_adds="false" # We currently don't want this enabled on non-Arch
...
+      _distro="Debian"
+    elif [ "$_distro_index" = "2" ]; then
+      _distro="Fedora"
+    elif [ "$_distro_index" = "3" ]; then
+      _distro="Suse"
+    elif [ "$_distro_index" = "4" ]; then
+      _distro="Ubuntu"
+    else
+      echo "Wrong index. aborting..."
+      exit 0
```
CodingKoopa, line 70: Rather than bailing out, this should prompt again for the distribution, like [here](https://github.com/Frogging-Family/linux-tkg/blob/9c3c155b3ed07a4aa04ec10d6fc0e8b5710a2f07/linux57-tkg/linux57-tkg-config/prepare#L824-L836).

linux57-tkg/install.sh
```diff
@@ -44,17 +47,46 @@ _misc_adds="false" # We currently don't want this enabled on non-Arch
...
-    sudo apt install git build-essential kernel-package fakeroot libncurses5-dev libssl-dev ccache bison flex
+    sudo apt install git build-essential kernel-package fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev -y
+  elif [ "$_distro" = "Fedora" ]; then
+    msg2 "Installing dependencies"
+    sudo dnf install fedpkg fedora-packager rpmdevtools ncurses-devel pesign grubby qt5-devel libXi-devel gcc-c++ git ccache flex bison elfutils-libelf-devel openssl-devel dwarves rpm-build -y
+  elif [ "$_distro" = "Suse" ]; then
+    msg2 "Installing dependencies"
+    sudo zypper install -y rpmdevtools ncurses-devel pesign libXi-devel gcc-c++ git ccache flex bison elfutils libelf-devel openssl-devel dwarves make patch bc rpm-build libqt5-qtbase-common-devel libqt5-qtbase-devel lz4
   else
     msg2 "Dependencies are unknown for the target linux distribution."
```
CodingKoopa, line 91: Since you've added the check for `$_distro` being one of the valid choices above, this code path shouldn't be reachable, and can probably be removed.

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
 
-  if [ "$_distro" = "Ubuntu" ]; then
+  if [ "$_distro" = "Ubuntu" ]  || [ "$_distro" = "Debian" ]; then
+
     if make -j ${_thread_num} deb-pkg LOCALVERSION=-${_kernel_flavor}; then
       msg2 "Building successfully finished!"
+
+      cd "$_where"
+
+      # Create rpms folder if it doensn't exist
```
CodingKoopa, line 196: `doesn't`

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
+
+    if make -j ${_thread_num} rpm-pkg EXTRAVERSION="_${_kernel_flavor}"; then
+      msg2 "Building successfully finished!"
+
+      cd "$_where"
+
+      # Create rpms folder if it doensn't exist
+      if ! [ -d RPMS ]; then
+        mkdir RPMS
+      fi
```
CodingKoopa, line 230: This can be shortened to `mkdir -p RPMS`.

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
+      #Clean up the original folder, unneeded and takes a lot of space
+      rm -rf ~/rpmbuild/
+
+      read -p "Do you want to install the new Kernel ? y/[n]: " _install
+      if [ "$_install" = "y" ] || [ "$_install" = "Y" ] || [ "$_install" = "yes" ] || [ "$_install" = "Yes" ]; then
+        
+        _kernelname=$_basekernel.${_kernel_subver}_$_kernel_flavor
+        _headers_rpm="kernel-headers-${_kernelname}*.rpm"
+        _kernel_rpm="kernel-${_kernelname}*.rpm"
+        _kernel_devel_rpm="kernel-devel-${_kernelname}*.rpm"
```
CodingKoopa, line 244: Would it be possible to get the name of the RPMs without using globbing? I'm concerned about globbing being able to capture more than one filename.

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
+        
+        if [ "$_distro" = "Fedora" ]; then
+          sudo dnf install RPMS/$_headers_rpm RPMS/$_kernel_rpm RPMS/$_kernel_devel_rpm
+        elif [ "$_distro" = "Suse" ]; then
+          msg2 "Some files from 'linux-glibc-devel' will be replaced by files from the custom kernel-hearders package"
+          msg2 "To revert back to the original kernel headers do 'sudo zypper install -f linux-glibc-devel'" 
+          sudo zypper install --replacefiles --allow-unsigned-rpm RPMS/$_headers_rpm RPMS/$_kernel_rpm RPMS/$_kernel_devel_rpm
+        fi
```
CodingKoopa, line 252: Perhaps `cd RPMS` to make these commands a little shorter.

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
+        
+        msg2 "Install successful" 
       fi
     fi
   fi
+
```
CodingKoopa, line 258: Unneeded newline (sorry :3).

bkdwt:
Hey @AdelKS, there's something wrong with 5.8 script: https://pastebin.com/y7F2T9gu

linux57-tkg/install.sh
```diff
@@ -132,74 +167,118 @@ if [ "$1" = "install" ]; then
...
+    msg2 "To uninstall a version, you should remove the linux-image, linux-headers and linux-libc-dev associated to it (if installed), with: "
+    msg2 "      sudo apt remove linux-image-VERSION linux-headers-VERSION linux-libc-dev-VERSION"
+    msg2 "       where VERSION is displayed in the lists above, uninstall only versions that have \"tkg\" in its name"
+  elif [ "$_distro" = "Fedora" ]; then
+    dnf list --installed kernel*
+    msg2 "To uninstall a version, you should remove the kernel, kernel-headers and kernel-devel associated to it (if installed), with: "
+    msg2 "      sudo dnf remove --noautoremove kernel-VERSION kernel-devel-VERSION kernel-headers-VERSION"
+    msg2 "       where VERSION is displayed in the second column"
+  elif [ "$_distro" = "Suse" ]; then
+    zypper packages --installed-only | grep kernel.*tkg
```
CodingKoopa, line 278: `kernel.*tkg` should be quoted for consistency with the other `grep`s.

CodingKoopa:
> Hey @AdelKS, there's something wrong with 5.8 script: https://pastebin.com/y7F2T9gu

It looks like `$pkgver` is not set [here](https://github.com/Frogging-Family/linux-tkg/blob/cdbf5273ad364cef1d9ccec94e40ceb0de334121/linux58-tkg/linux58-tkg-config/prepare#L125-L126), as it is exclusive to the Arch PKGBUILD. I guess something about how this PR sets `_distro="Arch"` is wrong.

Kodehawa:
image

The 5.8 script doesn't seem to work properly in Fedora either. Odd.

Kodehawa:
Even after setting _distro properly in install.sh, it still doesn't work:
image

AdelKS:
Sorry @Kodehawa @bkdwt I made a mistake, I fixed it now x)

linux57-tkg/install.sh
AdelKS: Laziness had struck me there xD Okay will do!

linux57-tkg/install.sh
AdelKS: oh right, that reminds me: I completely forgot about the other distros, they should be able to do `./install.sh config` without having to worry about giving out their distro type. I will fix that

linux57-tkg/install.sh
AdelKS: The compilation adds for example `+-1.x86_64` at the end of the package names that is different if RPM or DEB, it shouldn't blob more that one file since the `$_kernelname` variable contains the kernel version + `tkg-pds` or `tkg-${customname}` if the user gave one. Though I think it is possible to write the filename exactly.

linux57-tkg/install.sh
AdelKS: yes!

linux57-tkg/install.sh
AdelKS: It was done in purpose actually! I like to have a free space after the `if` condition and before the `fi` closing keyword. But I agree here in particular it kills the nice nesting.

linux57-tkg/install.sh
AdelKS: done!

linux57-tkg/install.sh
AdelKS: thanks!

linux57-tkg/install.sh
AdelKS: thanks!

Kodehawa:
image
This is still an issue, so it won't ever built :<

AdelKS:
Very weird, I tried the entire install process on my Fedora VM and didn't have any issue, can you attach your `customization.cfg` file ?

Kodehawa:
Actually I just found the actual error, lol
```
kernel/sched/pelt.c:413:5: error: redefinition of 'update_thermal_load_avg'
  413 | int update_thermal_load_avg(u64 now, struct rq *rq, u64 capacity)
      |     ^~~~~~~~~~~~~~~~~~~~~~~
In file included from kernel/sched/pelt.c:29:
kernel/sched/pelt.h:21:1: note: previous definition of 'update_thermal_load_avg' was here
   21 | update_thermal_load_avg(u64 now, struct rq *rq, u64 capacity)
      | ^~~~~~~~~~~~~~~~~~~~~~~
kernel/sched/pelt.c: In function 'update_thermal_load_avg':
kernel/sched/pelt.c:415:33: error: 'struct rq' has no member named 'avg_thermal'
  415 |  if (___update_load_sum(now, &rq->avg_thermal,
      |                                 ^~
kernel/sched/pelt.c:419:25: error: 'struct rq' has no member named 'avg_thermal'
  419 |   ___update_load_avg(&rq->avg_thermal, 1);
      |                         ^~
make[4]: *** [scripts/Makefile.build:281: kernel/sched/pelt.o] Error 1
make[3]: *** [scripts/Makefile.build:497: kernel/sched] Error 2
make[2]: *** [Makefile:1756: kernel] Error 2
make[2]: *** Waiting for unfinished jobs....
```

Seems like BMQ doesn't work properly either? Odd. Lemme re-try with CFS.
Attaching configuration.cfg anyway
[Uploading customization.cfg…]()

Kodehawa:
image
Also 5.9 doesn't work because the branch isn't called 5.9.y yet. It's an -rc.

AdelKS:
The customization file you attached works.

For the 5.9, I will check if there's an RC branch :)

Kodehawa:
> The customization file you attached works.
> 
> For the 5.9, I will check if there's an RC branch :)

Yeah, had to re-clone the repo. It was on a cursed state probably :P

Kodehawa:
image
kernel-devel replaces the current kernel-devel instead of being installed alongside it (like is meant to). It also installs kernel-debug-devel from @updates, for whatever reason.

AdelKS:
It will demand a little bit of work to make the script work with linux RC versions. I won't do it. Maybe I will just disable the script for now till release. What do you think @CodingKoopa @Tk-Glitch ?

So, otherwise, does dkms work properly now @Kodehawa ? I don't know why it installs `debug-devel` but why not. About `kernel-devel` replacing the current one, I couldn't figure out how to do that. If you can help, you can tell the script to not install the packages and do it by yourself and see if you can find a good command :)

Kodehawa:
Yeah. All seems to be working fine, including DKMS. 
The only issue is kernel-devel being replaced (which sort-of is an actual issue, because that could end up in the old kernel not being able to build dkms modules).

AdelKS:
I will see what I can do, but at least it's working. Which is an improvement from not existing at all for Fedora. You can do the hacky way and install the needed packages for the kernel you are running before installing modules.

linux57-tkg/install.sh
CodingKoopa: Ah, that's rather annoying. I suppose this is fine as is, as matching the exact filename sounds like a pain, and, thinking about it more, if it were to glob more than one file somehow, it wouldn't be the end of the world.

CodingKoopa:
> It will demand a little bit of work to make the script work with linux RC versions. I won't do it. Maybe I will just disable the script for now till release.

What exactly is breaking RC versions for non-Arch platforms?

AdelKS:
I would have to compile for Fedora and Ubuntu to see the naming scheme with the RC to blob it properly. And I need to use `master` branch from git instead of a `5.9.y` one that doesn't exist. I may end up implementing it, I've had enough for now x)

linux57-tkg/install.sh
AdelKS: Yea I think it is okay, and the script uses the distro's package manager, that shows what's about to be installed and asks for confirmation. We'll see if someone opens an issue about that, I will figure something out at that time :P

bkdwt:
@AdelKS 
Screenshot from 2020-08-31 18-41-50

Now the 5.8 version is working on OpenSUSE :100:

AdelKS:
Awesome!

bkdwt:
@AdelKS 5.8tkg compiled with only support for my hardware gives me more smooth frametimes with zero stutter and lag I have on 5.7tkg on Shadow of Tomb Raider. Even when my FPS drops bellow 60, I feel the game running smoother unlike 5.7, where my GPU was oscillating between 0 and 100%.

Thank you very much again for your hard work and for thanks indeed for @CodingKoopa. 🙏🏻

bkdwt:
Maybe is time to merge to master branch? 🤔

AdelKS:
Yea @Tk-Glitch will have a look at it when he will have time

iWeaker4you:
It gives me problems with akmod, I have problems loading it, and it ends up using nouveau, I don't know if the problem is in the debug kernel ... I need to investigate well.

AdelKS:
I got the same issue @iWeaker4you with Ubuntu 20.04, but the kernel 5.7 worked well for me. How about you ? give a try to 5.7.

iWeaker4you:
@AdelKS  is so strange 'cause on Ubuntu for me is working well kernel 5.8 with Nvidia Graphic Cards,  i'll give a try to 5.7 on Fedora.
Captura de pantalla de 2020-09-10 12-58-04

AdelKS:
@iWeaker4you I will retry a compile on 5.8 for Ubuntu with Nividia.

linux57-tkg/install.sh
```diff
@@ -132,74 +168,115 @@ if [ "$1" = "install" ]; then
...
-        if ! { [ -f installed-kernels ] && grep -Fxq "$_kernelname" installed-kernels; }; then
-          echo $_kernelname >> installed-kernels 
-        fi   
+      cd "$_where"
+
+      # Create RPMS folder if it doesn't exist
+      mkdir -p RPMS
+      
+      # Move rpm files to RPMS folder inside the linux-tkg folder
+      mv ~/rpmbuild/RPMS/x86_64/* "$_where"/RPMS/
```
gitfib, line 231: ~/rpmbuild is the default %{_topdir} so it might also be used for other packages

linux57-tkg/install.sh
```diff
@@ -132,74 +168,115 @@ if [ "$1" = "install" ]; then
...
+
+      #Clean up the original folder, unneeded and takes a lot of space
+      rm -rf ~/rpmbuild/
```
gitfib, line 234: ~/rpmbuild is the default %{_topdir} so it might also be used for other packages

gitfib (commented):
The script falsely assumes the user doesn't use ~/rpmbuild for other packages, which is possible because it's the default %{_topdir}

linux57-tkg/install.sh
AdelKS: You mean other built packages right ? I will change it so it only moves tkg related packages then.

AdelKS: Sorry for the late answer I thought this comment was for a new PR that wasn't mine x) I will take care of it in my currently open PR.

gitfib: Thanks, I did find it weird that I could review a merged PR. Yeah, other built packages and the removal of them. I'm not really sure how to specify a different _topdir in make

AdelKS: You are welcome! I did the change, you can review in #127

PR #62, opened 2020-09-09
Add initial -rc support to the install.sh script.

### LINKED ISSUE #58: Wrong git branch used for 5.9RC
The RC for 5.9 lives in branch "master" https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
This causes an error when the script tries to clone the branch linux-5.9.y
The install script should clone the master branch if the user wants to compile the RC

Line 114 in the install script reads
`git clone --branch linux-$_basekernel.y --single-branch --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git linux-${_basekernel}`

could be

`git clone --branch master --single-branch --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git linux-${_basekernel}`

Kodehawa:
Fixes #58 

Marginally tested. Builds, creates the RPMs and works fine when building.
Also installs the kernel fine.

Please someone test in Debian, don't have any debian-based machine to test :(
Confirmed to work in Fedora.

Fully reusable for future -RCs.

### DIFF
--- a/linux59-rc-tkg/install.sh
+++ b/linux59-rc-tkg/install.sh
@@ -21,6 +21,8 @@ set -e
 
 _where=`pwd`
 srcdir="$_where"
+# This is an RC, so subver will always be 0
+_kernel_subver=0
 
 source linux*-tkg-config/prepare
 
@@ -100,7 +102,9 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   if [ -d linux-${_basekernel}-${_sub}.orig ]; then
     rm -rf linux-${_basekernel}-${_sub}.orig
   fi
-
+  
+  msg2 "Building version ${_basekernel}.${_kernel_subver}_${_sub}"
+  
   if [ -d linux-${_basekernel}-${_sub} ]; then
     msg2 "You already have the current -rc .tar.gz downloaded, cleaning up and re-downloading..."
     rm -r $_where/linux-${_basekernel}-${_sub} 
@@ -227,7 +231,7 @@ if [ "$1" = "install" ]; then
 
       read -p "Do you want to install the new Kernel ? y/[n]: " _install
       if [ "$_install" = "y" ] || [ "$_install" = "Y" ] || [ "$_install" = "yes" ] || [ "$_install" = "Yes" ]; then
-        _kernelname=${_basekernel}_${_sub}_$_kernel_flavor
+        _kernelname=${_basekernel}.${_kernel_subver}_${_sub}_$_kernel_flavor
         _headers_rpm="kernel-headers-${_kernelname}*.rpm"
         _kernel_rpm="kernel-${_kernelname}*.rpm"
         _kernel_devel_rpm="kernel-devel-${_kernelname}*.rpm"
--- a/linux59-rc-tkg/install.sh
+++ b/linux59-rc-tkg/install.sh
@@ -30,7 +30,7 @@ source customization.cfg
 
 if [ "$1" != "install" ] && [ "$1" != "config" ] && [ "$1" != "uninstall-help" ]; then
   msg2 "Argument not recognised, options are:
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Nothing jumping to my face regarding potential breakage on Debian/buntu. A confirmation would be nice but otherwise I'm fine with this for RCs only. It's not flexible enough for an unified script though, so be aware it might get changed quite a bit at some point. If you're okay with that I'm fine with merging.

AdelKS:
I will report back tonight!

Also, is there a way to reset the kernel's files without having to delete the entire folder and unpack the tar again ? That's very intensive on the disk. I actually got an idea on how to keep on using Git for the RC by simply checking out the RC tag, and it's written in the `prepare` file, in the `_sub` variable.

Kodehawa:
> I will report back tonight!
Thanks :3

> Also, is there a way to reset the kernel's files without having to delete the entire folder and unpack the tar again ? That's very intensive on the disk.
To be fair, the RC is fairly short-lived, so this would still happen once a week, but we could create a makeshift git repo and stash changes after we patch everything and build it then discard the stash on the exit cleanup, leaving the folder in a clean state.

> I actually got an idea on how to keep on using Git for the RC by simply checking out the RC tag in the master branch, and it's written in the prepare file, in the _sub variable.
Probably a better idea. I'll look into it.
I just tried to make it work on my machine to test something, though I'll admit it took way too much time because compiling the kernel on my lappy took incredibly long (to test if it worked fine) haha

AdelKS:
Oh the temporary git repo is a very smart idea!

I'll look into Git too after trying out the code as-is.

Also, I have a tip for you to test things faster! Use modprobed-db so it uses localmodconfig, the compilation will take you few minutes 😁

Kodehawa:
> Also, I have a tip for you to test things faster! Use modprobed-db so it uses localmodconfig, the compilation will take you few minutes 

I used it, but building RPMs in fedora took like 20 minutes per run lmao

> PS: actually now that I think more about the temporary local git repo, it will need to `git add` every single file in the folder, I don't know how slow that is.

`git add -A` isn't too slow. Way faster than the way we're doing it rn, at least.

Thanks <3

AdelKS:
Sorry didn't take the time to do it tonight xD Tomorrow I promise! And I will give you a patch with the Git approach on top of it :D

About the `git stash`, I think the current approach in `install.sh` is better with `git reset --hard HEAD` and `git clean -f -d -x` because it doesn't even go through moving files to stash!

Kodehawa:
Hmm. Yeah, you're right. 

No issue!

AdelKS:
I started compiling for Ubuntu, there's already a double `rc4` appended to the DEBS, can you add me as a contributor to your fork so I can push directly the small fixes ? :D I can do them in a separate branch so you can check them out and merge them in master if you wish

Kodehawa:
> I did a first try, there's a double `rc4` appended to the DEBS so it fails at that already, can you add me as a contributor to your fork so I can push directly the small fixes ? :D I can do them in a separate branch so you can check them out and merge them in master if you wish

Nice one. On Fedora it doesn't append the rc4 so I had to append it, had no idea about deb.
I'll add you, push to master if you wanna haha

AdelKS:
Thanks! Done! I can push now :D

AdelKS:
Okay I made the changes I told you about! Tested them on both Ubuntu and Fedora and it works. It wasn't that much work actually, when compared to the original script. I don't know anymore why I didn't want to implement the RC script. 

This fixes #58 ! And I think it's now good for merge :D

Kodehawa:
Tested it myself, works fine! Pretty sure we can merge now.

> PS: the stage of making the actual RPM packages is so slow... I can't believe they didn't multi-thread it

Literally slower than building the kernel :<

AdelKS:
They need to step up their rpm-pkg game! Because it's ridiculous...

@Tk-Glitch 
One merge please - frog

Tk-Glitch:
All good! Thank you guys :heart: :frog:

PR #65, opened 2020-09-12
Fix cleanup step on non-arch distros

### LINKED ISSUE #64: `exit_cleanup()` at end of `./install.sh config` breaks the source tree
I had trouble trying to build linux58-tkg on Debian with:

```
cd linux58-tkg &&
./install config &&
cd linux-5.8 &&
make bindeb-pkg
```

(The last line is how I build all my kernels normally.)

Build failed due to missing headers and files. Which files specifically were missing isn't relevant here, because they're obviously deleted by `exit_cleanup()` in `linux58-tkg/linux58-tkg-config/prepare` at the end of `./install.sh config`.

The cleanup on exit is apparently done with the assumption that the compilation itself has finished by that point, but that's not true when `install.sh` is used only for patching and configuring.

AdelKS:
This PR fixes #64 

The `install.sh` script cleans up by itself what's in the linux git folder and does not need cleaning up. And it was affecting the `config` feature it offered of patching&configuring the git folder and leaving it in that state so the user can compile it in the way he likes.

### DIFF
--- a/linux57-tkg/linux57-tkg-config/prepare
+++ b/linux57-tkg/linux57-tkg-config/prepare
@@ -917,60 +917,60 @@ exit_cleanup() {
     rm -f "$_where"/"$_p"
   done
 
-  if [ "$_NUKR" = "true" ] && [ "$_where" != "$srcdir" ]; then
-    rm -rf "$_where"/src/*
-    # Double tap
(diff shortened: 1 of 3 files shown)


AdelKS:
@Tk-Glitch 
One merge please -frog v2

PR #67, opened 2020-09-13
Draft: Use latest prjc-pds code from alfred chen's linux-5.8.y-prjx-pds branch

AdelKS:
Hello!

I discovered that Alfred Chen is working on a separate branch for PDS and it has for now one more commit than the main branch. I don't know for how long the branch will be used and if it is worth following. But I opened this PR just to notify about it.

I switched to Project-C PDS to find any problem/regression and help out in its development.


Tk-Glitch:
r3 was released and merged with https://github.com/Frogging-Family/linux-tkg/commit/41cfeabff53e11604976efcd30e02841b889f478

PR #68, opened 2020-09-14
Fix build fail certificate UEFI on Debian 5.7/5.8/5.9rc

iWeaker4you:
Reason that the compilation fails due to the need for this certificate, it is not necessary and only causes problems

### DIFF
--- a/linux57-tkg/install.sh
+++ b/linux57-tkg/install.sh
@@ -140,6 +140,7 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   cp /boot/config-`uname -r` .config
   if [ "$_distro" = "Debian" ]; then #Help Debian cert problem.
     sed -i -e 's#CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/test-signing-certs.pem"#CONFIG_SYSTEM_TRUSTED_KEYS=""#g' .config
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Blindly merging thanks to Debian's stability :tm: 
Thanks! :+1:

AdelKS:
I think these are hacky fixes : I mean they are there for a reason right ? xD Though I have no idea how to make the script abide by them / regenerate them for custom kernels.

iWeaker4you:
@AdelKS  I don't know why, but in the Debian kernel compilation guides, they recommend disabling it and commenting on it as it complicates compilation.
And since I have a 100% UEFI machine, that problem appeared to me, and I used the same method and it is currently working fine.

Captura de pantalla de 2020-09-14 08-32-11

AdelKS:
I see! Well if they say so in their guide... The important part is to have it work! :D I see that you use BMQ, any particular reason ?

iWeaker4you:
@AdelKS  No haha, when I was compiling I thought it was due to the PDS patch and I selected BMQ to discard, but when I saw that it was the certificate I forgot to change it, at night I compiled it again. XD

AdelKS:
haha got it! On Ubuntu I was using modprobed-db and it was working like a charm. You should give it a try! It divides the compile time by 10.

PR #78, opened 2020-09-30
linux58-tkg: Remove "include/linux/compiler*.h: define asm_volatile_g…

imaami:
…oto"

It was merged upstream in commit 8bd66d147c88bd441178c7b4c774ae5a185f19b8
in Linux v5.0, but was never removed from glitched-base. Applying it just
just adds a duplicate ifndef-define-endif block that has no effect.

Signed-off-by: Juuso Alasuutari <juuso.alasuutari@gmail.com>


Tk-Glitch:
Thanks! I'll regen the hashes on my end.

PR #80, opened 2020-10-03
linux58-tkg: Clean up and fix glitched-base patch

imaami:
Split most changes into properly attributed patches from their respecive
source repos, taking into account where each one seems to have been first
introduced. The attribution was done manually on a best-effort basis by
inspecting git logs from repositories such as tkg, zen, xanmod, and pf.

This commit also fixes a case of an invalid macro effectively disabling the
"defer+madvise" patch. Glitched-base included a variant of the patch where
the change depends on a config option named CONFIG_AVL_INTERACTIVE. This
config option isn't even mentioned anywhere else in linux-tkg, so the patch
has likely never been active in any build. I replaced the patch with an
identical one from zen-kernel and modified it to depend on CONFIG_ZENIFY.

Signed-off-by: Juuso Alasuutari <juuso.alasuutari@gmail.com>


Tk-Glitch:
Thank you imaami!

PR #84, opened 2020-10-11
More dependencies added and Kernel package will be installed via Wget on Debian Linux

iWeaker4you:
Kernel package is not found in the official Debian repos, especially in Debian 11 Testing, with that it will be installed as well as more dependencies.

### DIFF
--- a/linux57-tkg/install.sh
+++ b/linux57-tkg/install.sh
@@ -81,9 +81,16 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
     exit 0
   fi
 
-  if [ "$_distro" = "Ubuntu" ] || [ "$_distro" = "Debian" ]; then
+  if [ "$_distro" = "Ubuntu" ]; then
     msg2 "Installing dependencies"
     sudo apt install git build-essential kernel-package fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev -y
+  elif [ "$_distro" = "Debian" ]; then
+    msg2 "Installing dependencies"
+    sudo apt install git build-essential po-debconf gettext bc xmlto m4 libsigsegv2 fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev docbook-xsl libxml2-utils xsltproc intltool-debian  libcroco3 libncurses-dev rsync libelf-dev -y
+    TEMP_DEB="$(mktemp)" &&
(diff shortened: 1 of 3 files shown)


linux57-tkg/install.sh
```diff
@@ -81,9 +81,16 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
...
 
-  if [ "$_distro" = "Ubuntu" ] || [ "$_distro" = "Debian" ]; then
+  if [ "$_distro" = "Ubuntu" ]; then
     msg2 "Installing dependencies"
     sudo apt install git build-essential kernel-package fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev -y
+  elif [ "$_distro" = "Debian" ]; then
+    msg2 "Installing dependencies"
+    sudo apt install git build-essential po-debconf gettext bc xmlto m4 libsigsegv2 fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev docbook-xsl libxml2-utils xsltproc intltool-debian  libcroco3 libncurses-dev rsync libelf-dev -y
+    TEMP_DEB="$(mktemp)" &&
+    wget -O "$TEMP_DEB" 'http://archive.ubuntu.com/ubuntu/pool/universe/k/kernel-package/kernel-package_13.018+nmu2_all.deb' &&
```
h3o66, line 91: imho I would get the package for debian from the debian backports instead and not mix this up if this is needed.
https://packages.debian.org/de/stretch-backports/kernel-package

h3o66: So after checking this a bit more, the `kernel-package` should not be needed anymore.
From a forum:
http://forums.debian.net/viewtopic.php?f=10&t=133625

From the offical debian kernel build page:
https://wiki.debian.org/BuildADebianKernelPackage

iWeaker4you: I will make the change to that backport, Linux TKG does not compile if it does not have this package installed, it is a requirement, and if I read that forum, it is one of the reasons that it is no longer available or will not be available in version 11 of Debian .

h3o66: Well the latest version of the package is in Debian 9 in this case.

For ubuntu this package is still there:
https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=kernel-package

h3o66: Also from the filelist and the commands what are used at least i do not see why it should be needed anymore.

Filelist of the `kernel-package`:
https://packages.debian.org/de/stretch-backports/all/kernel-package/filelist

PR #94, opened 2020-10-25
Fedora 33 support.

Aeirex:
This makes the install.sh work with a fresh install of Fedora 33. Make is also missing from Fedora install which all versions do not come with and will make the build fail.

Some notes:
- Fedora 32 and below will not work, because they require qt5-devel which the package is now installed as a different package name in Fedora 33. I did not have to include anything to make it build as the qt5 was automatically installed by default or by the other build tools which were pulled.
- I did kernel.spec of kernel tree (kernel-5.9.0-36.eln104.src) from RedHat to get these dependencies required for compiling. I did tests to make sure they are as minimal as possible.
- This is just to help with understanding the changes of packages to upstream to the release of Fedora 33, as we probably should have include qt5-devel package for backward compatibly with older versions of Fedora.


AdelKS:
Dropping support for Fedora 32 is a bit problematic since not everyone will switch to it. Maybe there's a way to tell which version is being run with something like

```shell
source /etc/*-release
if [ "${DISTRIB_RELEASE}" -le "32" ]; then
  #...
fi
```

aukuste:
There is a way to tell what fedora version you're running.
`rpm -E %fedora`
Which outputs 32 for me.

PR #95, opened 2020-10-26
linux-tkg: cosmetics in localversion if using llvm

ebiroud:
There is one additional "-" which shouldn't be here.


Tk-Glitch:
Indeed. Thanks!

PR #99, opened 2020-10-27
Add patch for Bug 209745, FS#68346

thotypous:
This adds a patch for a usability-crippling Bluetooth issue which is affecting Intel Bluetooth adapters since 5.9.0.

See:
* https://bugzilla.kernel.org/show_bug.cgi?id=209745
* https://bugs.archlinux.org/task/68346
* https://bbs.archlinux.org/viewtopic.php?pid=1932543


Tk-Glitch:
Thanks for the heads up! Your PR doesn't allow the patch to apply, so I have decided to add it to base patchset.

PR #100, opened 2020-10-27
Add back #62 Add initial -rc support to the install.sh script.

Hyper-KVM:
This PR #62  was merged after starting the work on the unified build script.

Also add 5.10 RC to opt_ver case.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -27,6 +27,11 @@ source linux-tkg-config/prepare
 # Run init script that is also run in PKGBUILD, it will define some env vars that we will use
 _tkg_initscript
 
+if [[ "$_sub" = rc* ]]; then
+  # if an RC version, subver will always be 0
+  _kernel_subver=0
+fi
+
 case "$_basever" in
 	"54")
 	opt_ver="4.19-v5.4"
@@ -38,6 +43,9 @@ case "$_basever" in
 	opt_ver="5.8%2B"
 	;;
 	"59")
+	opt_ver="5.8%2B"
+	;;
+	"510")
 	opt_ver="5.8%2B"
 	;;
 esac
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thank you! :frog: :heart:

AdelKS:
Super! I will start working on unifying all the linux git folders into one and just `git chekout` between shallow branches. @Tk-Glitch what do you think of actually using the git folder instead of tarballs for Arch's build script ?

Tk-Glitch:
I'm still against it as the default. I'd be okay with it as an optional path if it's not too invasive :frog: .

AdelKS:
I will try to think of it. I keep on suggesting this because the downside of the tarballs is disk usage for SSDs: the kernel is decompressed every time `makepkg` is run right ?

Tk-Glitch:
Yes I know :stuck_out_tongue: , but you can easily use a tmpfs to do that in RAM if you don't have a HDD or if you're worried about your SSD's lifespan. Makepkg has options for that. Of course if you don't have enough RAM for the task it's more involved :frog:

Tk-Glitch:
Screenshot_20201027_201856
By the way, here's the tmpfs usage for a fresh 5.10rc1 build. A 4GB RAM equipped system would do just fine here, while building faster than on SSD, and without the disk writes.

AdelKS:
I was about to actually ask you how much tmpfs is needed. 4GB that's little! I will set that up then. Does it have to be permanent, in the `fstab` ?

Tk-Glitch:
Systemd will create a default tmpfs in /tmp with a size of half your total RAM if it doesn't exist in fstab. You can override that by adding an entry to your fstab to set it to whatever size you want. If you fill up your RAM completely for some reason, swap will then be used.

See https://wiki.archlinux.org/index.php/Tmpfs

AdelKS:
Thanks for the information!

PR #101, opened 2020-10-27
install.sh: Fix LLVM build restart config

Hyper-KVM:
Fix MuQSS illegal character uppercase in source package name for .deb build
Fix _cpusched not being sourced for package name
Fix llvm not in package name when using llvm

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -196,11 +196,19 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   if [ "$_distro" = "Debian" ]; then #Help Debian cert problem.
     sed -i -e 's#CONFIG_SYSTEM_TRUSTED_KEYS="debian/certs/test-signing-certs.pem"#CONFIG_SYSTEM_TRUSTED_KEYS=""#g' .config
   fi
-  yes '' | make oldconfig
+  yes '' | make ${llvm_opt} oldconfig
   msg2 "Done"
 
   # apply linux-tkg patching script
   _tkg_srcprep
+
+  # source cpuschedset since _cpusched isn't set
+  source "$srcdir"/cpuschedset
+
+  # Uppercase characters are not allowed in source package name for debian based distros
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -196,12 +196,20 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
...
   msg2 "Done"
 
   # apply linux-tkg patching script
   _tkg_srcprep
 
+  # source cpuschedset since _cpusched isn't set
+  source "$srcdir"/cpuschedset
+
+  # Uppercase characters are not allowed in source package name for debian based distros
+  if [ "$_cpusched" = "MuQSS" ] && [ "$_distro" = "Debian" ] || [ "$_distro" = "Ubuntu" ]; then
```
Tk-Glitch, line 209: This is equivalent to:
```
if ( [ "$_cpusched" = "MuQSS" ] && [ "$_distro" = "Debian" ] ) || [ "$_distro" = "Ubuntu" ]; then
```
Which isn't what we want, is it?

I guess we want:
```
if [ "$_distro" = "Debian" ] || [ "$_distro" = "Ubuntu" ] && [ "$_cpusched" = "MuQSS" ]; then
```

CodingKoopa, line 209: I believe this evaluates as:
```bash
(_cpusched == "MuQSS" && _distro == "MuQSS") || _distro == "Ubuntu"
```
If so, this code would have the unintended consequence of always setting `_cpusched="muqss"` on Ubuntu. Since this codebase does seem to the allow the usage of `[[`, I might change this to:
```bash
  if [[ "$_cpusched" = "MuQSS" && ( "$_distro" = "Debian" || "$_distro" = "Ubuntu" ) ]]; then

CodingKoopa (changes requested):
:frog:

install.sh
CodingKoopa: Oh, it looks like @Tk-Glitch beat me to this, haha.

Hyper-KVM: oh yeah true lol

Hyper-KVM: ok fixed :)

Tk-Glitch:
Thanks! :frog: :heart:

PR #102, opened 2020-10-27
Add custom LLVM compiler path support

Hyper-KVM:


Tk-Glitch:
:frog:

AdelKS:
@Hyper-KVM Were you able to use Google's LLVM LTO and PGO patches ?

Hyper-KVM:
> @Hyper-KVM Were you able to use Google's LLVM LTO and PGO patches ?
I tried LTO but I seem to have a memory leak which makes it useless for me as it causes the system to lockup. I want to eventually get LTO support going too.

AdelKS:
Oh I see, thanks for the feedback. Yea that would be awesome. Google had some hold statements about the improvements of LTO + PGO for their kernels, something like the Kernel taking 20% less cycles on the CPU

PR #103, opened 2020-10-27
README.md: Add xbps-src (Void Linux) build optional configuration opt…

Hyper-KVM:
…ions customization.cfg and userpatches instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,11 @@ cd path/to/linux-tkg
 git clone -b tkg https://github.com/Hyper-KVM/void-packages/
 cd void-packages
 ./xbps-src binary-bootstrap
+# Optional: edit customization.cfg located in srcpkgs/linux-tkg/files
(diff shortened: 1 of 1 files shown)


AdelKS:
I think it is important to add a README.md at the root of this Git repository that explains what this repository is about. Because it's no written anywhere.

AdelKS:
Oh sorry I think I may have not explained myself properly. I meant that @Tk-Glitch or you in this PR for example, add a README.md at the root of this repository that explains what it's about. Since the idea is to switch to your changes only. Maybe add a short intro to what this repo is about in the README.md

Something like:

" linux-tkg is a script that downloads, patches and compiles the Linux Kernel from the official Linux Git sources, with extra patches that aim for better Desktop/gaming performance. 

The script has been designed so that it's easy for the user to chose which patches to use and also use his own patches: the file to edit for that matter is `customization.cfg`. More information is available on it."

Tk-Glitch:
Considering how different the Void path is, I can see this being useful.

PR #108, opened 2020-10-31
update Debian build dependencies in "install.sh"

termac:
Split Debian and Ubuntu build dependencies, as "kernel-package" is not
available, but also not needed in newer Debian releases.
Add some other dependencies dpkg-buildpackage was complaining about.
With these dependencies the tkg kernel builds in a freshly bootstrapped
Debian "Buster" container.


Tk-Glitch:
Thank you! :frog: :heart:

PR #109, opened 2020-10-31
source 'customization.cfg' earlier in 'install.sh'

termac:
source 'customization.cfg' before '_tkg_initscript' is run in
'install.sh'. Otherwise some variables set in 'customization.cfg' are
empty when '_tkg_initscript' is run and have to be supplied again on the
prompt.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -22,6 +22,8 @@ set -e
 _where=`pwd`
 srcdir="$_where"
 
+source customization.cfg
+
 source linux-tkg-config/prepare
 
 # Run init script that is also run in PKGBUILD, it will define some env vars that we will use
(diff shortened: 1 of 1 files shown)


termac:
I did not thoroughly test this and do not know if there are any side effects by moving this around. It works fine for me but maybe someone else can try this for their use case.

Mantas-2155X:
I had to do this as well, because the script kept on insisting that my distro setting is empty and automatically set itself to Arch when the configuration was set to Ubuntu.

AdelKS:
Yea we are on a bug-hunt phase after the merger with the unique script, I haven't taken the time to thoroughly test the changes and fix anything that appears.

PR #120, opened 2020-11-15
llvm: Fix kcsan prompt showing since we don't want kcsan

Hyper-KVM:
also move Void LibreSSL out of noccache check so it still runs if ccache
is set off
Thanks to @Kodehawa for pointing this out, since LLVM11 is on Arch now :frog:


Tk-Glitch:
Thanks!

PR #122, opened 2020-11-17
Option to enable Binder and Ashmem modules for use with Anbox

Sukid:
After kernel version 5.7, the dkms modules for Anbox no longer work, and the built in versions of these modules need to be enabled in the kernel. Other Android related projects might need this as well? Either way, easy enough to switch on. :)


CodingKoopa:
The `prepare` file in the root of the repo should probably be deleted.

Sukid:
I thought I deleted that already? oops.

Tk-Glitch:
It was added back with https://github.com/Frogging-Family/linux-tkg/pull/122/commits/2dfa10e2af09224eebce4891122240bd0b341680

Sukid:
I don't know why it took me so long to get around to actually deleting it but that is so weird, could have sworn I already had. Now it's gone though.

Tk-Glitch:
Thanks!

Issue #115, opened 2020-11-08
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2


hakavlad:
Cannot install on Debian 10: https://pastebin.com/Yw83ZzAj

Tk-Glitch:
```
make[4]: *** No rule to make target 'debian/certs/debian-uefi-certs.pem', needed by 'certs/x509_certificate_list'.  Stop.
```
That looks familiar for some reason..

PR #126, opened 2020-11-21
Debian UEFI cert compile problem

iWeaker4you:


AdelKS:
Hello! I am not a `sed` expert so I want to confirm with, are you basically emptying `CONFIG_SYSTEM_TRUSTED_KEYS`, aka setting it to `CONFIG_SYSTEM_TRUSTED_KEYS=""` ? I need to understand that because I will open a PR soon where all the `sed` commands will be replaced with a script from the linux source code itself, meant to edit the `.config` file in a more human readable way.

Tk-Glitch:
Thank you! This should fix https://github.com/Frogging-Family/linux-tkg/issues/115

Tk-Glitch:
@AdelKS This will need exceptions depending on kernel version/commit id to handle failures (which isn't an issue with sed), which will make it even more annoying to read imho.

AdelKS:
Then forget about it x) @Tk-Glitch

PR #127, opened 2020-11-21
install.sh: unified linux git folder across kernel versions

AdelKS:
Hello!

I finally got to do the merger across kernel versions in a single git folder, where each version is a shallow cloned branch. Shallow cloning only happens when the user asks for the kernel version.

I also corrected a little issue where doing `./install.sh uninstall-help` won't work if the user doesn't set the `_distro` variable in `customization.cfg`.

I have tested the changes on Fedora 32, Ubuntu 20.10 and 20.04, and on Archlinux, @Hyper-KVM can you test that Void still works ? :D


Tk-Glitch:
Regarding the block you're unsure about, it's needed for people wanting to use `_NUKR="false"`.
```
# [Arch specific] Set to anything else than "true" to limit cleanup operations and keep source and files generated during compilation.
# Default is "true".
_NUKR="true"
```

AdelKS:
Okay so I did well to add a `[ "$_distro" == "Arch" ]` and change the condition for the `trap exit_cleanup EXIT` ? If so I  will just remove those comments.

AdelKS:
Oh right, @Tk-Glitch, should I implement an option of Archlinux to use this folder as you suggested ? I actually prefer your current approach now that I have enough RAM to make the entire makepkg process in `tmpfs` (in RAM), 16GB of RAM is also enough if one uses `modprobed-db`. If you prefer to do it yourself, the git cloning script is in a function in the `prepare` file.

install.sh
```diff
@@ -224,10 +224,7 @@ if [ "$1" = "install" ]; then
       mkdir -p RPMS
       
       # Move rpm files to RPMS folder inside the linux-tkg folder
-      mv ~/rpmbuild/RPMS/x86_64/* "$_where"/RPMS/
-
-      #Clean up the original folder, unneeded and takes a lot of space
-      rm -rf ~/rpmbuild/
+      mv ~/rpmbuild/RPMS/x86_64/*tkg* "$_where"/RPMS/
```
gitfib, line 227: That works, albeit removing the tkg build files was a good idea 😝

OK I found that it's possible to use `RPMOPTS='--define "_topdir /tmp/test"' make rpm-pkg` to build in a directory that isn't ~/rpmbuild, so I think it would be better to use the previous mv and rm -rf lines, but in a tkg temporary directory inside ~/.cache

gitfib (changes requested):

install.sh
AdelKS: Okay thanks for the tip! I will implement that and get back to you :D

AdelKS: Is there a problem with using `/tmp` ? Since it's in RAM, I guess it's better ?

gitfib: `/tmp` might run out of space (on my computer it has 7.8G and it doesn't seem to be enough, although I can increase it because I have 16G of RAM). Some would prefer it (in Arch they would do it via /etc/makepkg.conf), perhaps we can add a config option if there isn't something more standardized

AdelKS: The config option makes sense actually, I will do that, I just had an issue too on my VM of 5GB RAM, I though the `rpmbuild` folder takes less than 2GB but apparently not.

AdelKS: Do you have any idea how to change the working folder with `deb-pkg` ?

gitfib: Let's see... is it currently building at the kernel directory in `/debian`? It looks hardcoded, starting from `/scripts/package/mkdebian` (`mkdir -p debian/source/`)

gitfib: I guess symlinking the debian directory before `make` would do the trick?

AdelKS: The RPMs find themselves in the same folder as the linux-tkg git folder, I don't know exactly where the temporary files are put.

AdelKS: You are saying that there's a temporary `debian` directory created in the linux-tkg git folder, and if we create first `make` will write inside it ?

linux-tkg-config/prepare
```diff
@@ -6,11 +6,119 @@ ver58=18
...
+
+    msg2 "Current branch: $(git branch | grep "\*")"
+    msg2 "Reseting files to their original state"
+    
+    git reset --hard HEAD
+    git clean -f -d -x
+  fi 
+
+  if [[ "$_sub" = rc* ]]; then
+    msg2 "Switching to master branch for RC Kernel"
```
flgx16, line 83: Afaik master is ahead of RC releases so maybe it could be split up to `rc` and `master`? Maybe using tags for it would be the solution

AdelKS: Yes the script is doing is: switch to master, shallow clone the past 3 weeks, then checkout the RC tag in that time period where it's expected to be. Do you have a better idea on how to handle it ?

flgx16: Sorry, overlooked that RC tag checkout

AdelKS: No worries!

install.sh
gitfib: Sorry for the late reply. You mean the RPMs don't stay at `?/rpmbuild/RPMS/x86_64/` after `make rpm-pkg`?
Yeah I saw it created a `debian` directory, I believe `make` would write into it without error if we create/symlink it ourselves before it does. I'm going to need a very minimalistic kernel config in order to fully test it myself heh

AdelKS: Oh no the RPMs do stay, I was only talking about debs. Don't worry I have everything setup to test it, I will try your hack and see :D

gitfib: Haha yeah. Ok cool, give it a try, check if the `debian` directory isn't deleted automatically after `make`

AdelKS: Nah it doesn't work, I don't see any `debian` directory being created, I will just do your first suggestion in `~/.cache` for now.

gitfib: That's odd. I can download the kernel, run `make deb-pkg` after `make menuconfig` and see the `debian` directory shortly after. Maybe it's deleted relatively quick on a deb distro? It contains files such as `control` and `rules`

AdelKS: Oh maybe, but while things are compiling, when compared to RPM compilation, the folder isn't there any longer. So it is not some kind of working directory where compilation temp files are kept. But maybe nothing really is kept. The compiler itself does use `/tmp` though.

gitfib: Oh I see. /tmp doesn't run out of space, so might have to resort to using `pyinotify` or something 😅

AdelKS: I am not sure I understood you, as far as I understand `/tmp` is on the RAM for systemd distros and can by default fill up to half the amount of the total RAM. So it can run out of space and I already got that error. The thing is, when I said "`/tmp` is used by the compiler" I meant for a small amount of temporary intermediary compilation files. So nothing much. Why would I need to resort to `pyinotify` ?

gitfib: I just find it weird that building for rpm needs gigabytes of storage to build and deb not

AdelKS: Yea I have noticed that  when I was testing, and honestly I don't know why it produces so many side products, I could do `bin-rpmpkg` which doesn't produces all these files but the kernel-devel package isn't made anymore, needed to compile nvidia modules for example :/

AdelKS:
@Hyper-KVM a little ping, to check if void is good with this :P

Hyper-KVM (changes requested):
> @Hyper-KVM a little ping, to check if void is good with this :P

I suggest just dropping lines 440-442, I can cd into the correct directory from the template.

linux-tkg-config/prepare
```diff
@@ -325,10 +436,9 @@ _tkg_srcprep() {
   fi
 
   # Void
-  if [ "$_distro" = "Void" ] && [[ "$_sub" = rc* ]]; then
-    cd ${wrksrc}/linux-${_rc_kern_ver}
-  elif [ "$_distro" = "Void" ]; then
-    cd ${wrksrc}/linux-${_kern_ver}
+
+  if [ "$_distro" = "Void" ]; then
+    cd ${wrksrc}/linux-src-git
```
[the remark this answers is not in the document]
AdelKS: Oh wait I thought you used the git folder, but actually you don't you ? You extract tarballs ? I will have a closer look

AdelKS:
@Hyper-KVM I thought you were using the cloned git folder, done by `install.sh` but actually not at all, since the install procedure doesn't involve it at all. So I reverted the changes. 

Something related to it: as I suggested to @Tk-Glitch, I think it would be nice to use `install.sh` as the common install script, that will run in the background the needed commands, for a more unified approach, at least in from the front end point of view. Though maybe it will make things more obscure.

AdelKS:
Okay I tested the 5.10 code path, everything worked without the need to touch anything. Great! Void Linux should be fine since I reverted the changes I did to its code paths.

Now then, @Tk-Glitch 

One merge please -frog v2

Tk-Glitch:
Let's go :) Thanks Adel! :frog: :heart:

AdelKS:
Always happy to contribute!

PR #129, opened 2020-11-23
Add anbox to readme

iiFir3z:
Figured you may wanna add this to the readme so people don't go ahead and make configuration changes if its already an option.


Tk-Glitch:
Indeed. Thank you!

PR #144, opened 2021-01-06
add cpio as make depend

SimpliFly03:
I got an error "make[1]: *** [kernel/Makefile:146: kernel/kheaders_data.tar.xz] Error 127" when compiling 5.10.4. According to https://bbs.archlinux.org/viewtopic.php?id=261876 "cpio" package is needed. Installing that fixed the problem.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -53,7 +53,7 @@ pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="http://www.kernel.org/"
 license=('GPL2')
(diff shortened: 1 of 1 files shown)


CodingKoopa:
This was added to the Arch package [here](https://github.com/archlinux/svntogit-packages/commit/871f3c01f5d406189a3fdb9ff14e91c3d71772a6#diff-37538beb61ff63edebbf735dfcf39e5d732f49183d6beb097169d971875ca422R12), for reference.

SimpliFly03:
@CodingKoopa From the link you provided perl, tar and xz was also added.

SimpliFly03:
Force push above caused by me trying to fix partial commit message result from accidental enter press.

Tk-Glitch:
:+1: Thanks !

PR #149, opened 2021-01-08
Fix sandboxing for security

BL4CKH47H4CK3R:


Tk-Glitch:
Since this option is enabled by default on stock Arch's kernel, it should be added as a .cfg toggle instead of being enforced.

PR #150, opened 2021-01-09
Blacklist $_path from installed config

CodingKoopa:
It was decided [here](https://github.com/Frogging-Family/linux-tkg/pull/49#issuecomment-677631969) that we would like to not include the build path in the built package. This PR upholds this by adding an exception for `$_path` to `customization-full.cfg`.

Cheers ^.^


Tk-Glitch:
Thank you! :frog:

PR #153, opened 2021-01-13
enable CONFIG_EXPERT=y by default for the futex2

OneOfOne:
The futex2 patch depends on it.


Tk-Glitch:
The defconfig is pulled from Arch so ideally we'd want to set the option from the prepare script instead so the change isn't lost on next sync.

pchome:
FYI: You could modify the futex2 patch to set `default y`.
Even enabling and immediately disabling CONFIG_EXPERT will keep some default expert-mode options enabled.

_I came here looking for alternative solutions, but you seems stepping through the same mine field._

PR #154, opened 2021-01-14
Get userpatches from version-specific subdirectories

openglfreak:


Tk-Glitch:
Thank you! :frog: :heart:

PR #164, opened 2021-01-19
do not urlencode gcc patch version and link

termac:
Wget can handle none urlencoded urls and the gcc patch was not applied
in some cases because the urlencoded file path was not properly handled
after downloading.

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -365,7 +365,7 @@ _tkg_srcprep() {
   elif [ "${_distro}" = "Void" ]; then
     tkgpatch="${wrksrc}/enable_additional_cpu_optimizations_for_gcc_v10.1+_kernel_v${opt_ver}.patch" && _tkg_patcher
   else
-    tkgpatch="$srcdir/enable_additional_cpu_optimizations_for_gcc_v10.1+_kernel_v${opt_ver}+.patch" && _tkg_patcher
+    tkgpatch="$srcdir/enable_additional_cpu_optimizations_for_gcc_v10.1+_kernel_v${opt_ver}.patch" && _tkg_patcher
   fi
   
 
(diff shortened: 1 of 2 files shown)


linux-tkg-config/prepare
```diff
@@ -361,11 +361,11 @@ _tkg_srcprep() {
   if [ "$_compiler_name" != "-llvm" ]; then
     msg2 "Applying graysky's cpu opts patch"
     if [ "${_distro}" = "Arch" ]; then
-      tkgpatch="$srcdir/enable_additional_cpu_optimizations_for_gcc_v10.1%2B_kernel_v${opt_ver}.patch" && _tkg_patcher
+      tkgpatch="$srcdir/enable_additional_cpu_optimizations_for_gcc_v10.1+_kernel_v${opt_ver}.patch" && _tkg_patcher
```
AdelKS, line 364: You should simply revert this line and it's all good! Archlinux doesn't follow what's in `install.sh`

AdelKS (approved):
The changes are good! Except one line that needs to be reverted, I commented on it.

linux-tkg-config/prepare
termac: But what is this line than for in the first place? From what I gathered "Arch" is also used as default / fallback. The "%2B" escape does not work with `_tkg_patcher`, why should it be reverted?

AdelKS: This line is used when the `prepare` script is called from the `PKGUILD` script. And the `PKGBUILD` script downloads the files with the exact same file name involving the `%2B`

termac: Thank You for explaining! I think I understand better now how it works for Arch. I reverted the line for Arch.

AdelKS:
PR #127 got merged, where I implemented this fix already, can you give it a try ? :D

termac:
Looks good on my end! Thank You.

PR #166, opened 2021-01-22
Update modprobed-db to new usage

AdelKS:
Hello!

After opening [an issue in modprobed-db's github](https://github.com/graysky2/modprobed-db/issues/15), I got informed of the new recommended usage that actually fixes the issue with modules that fail to load, thus end up not being in the custom built kernel. Now `make localmodconfig` receives an extra argument that points directly towards the database, recall isn't even needed. This approach is better than the old one in every aspect.

Adel


Tk-Glitch:
Let's give it a try

PR #174, opened 2021-02-04
5.10.13

flgx16:
Looks like the patch for the regression in iwlwifi is in 5.10.13, removed it from the misc additions patch


Tk-Glitch:
Thanks! :frog: :heart:

PR #184, opened 2021-02-23
add pahole dependency to headers on 5.11+

Snoop05:
after that it can be removed from nvidia-tkg


Tk-Glitch:
Thanks :frog: :heart:

PR #189, opened 2021-02-26
Enable apparmor kernel parameter

jollycar:
Since running kernel 5.11, I cannot run docker, because apparmor is not set as a kernel parameter.

See:
https://stackoverflow.com/questions/66073983/docker-stopped-working-with-error-message-oci-runtime-create-failed/66101032#66101032
https://wiki.archlinux.org/index.php/AppArmor#Custom_kernel
https://github.com/docker/for-linux/issues/1199


Tk-Glitch:
We are using the stock Archlinux config, which doesn't pass the option to make it default. It wasn't set on 5.10 or 5.9 either.

You have various options to do so for your setup. Either add `lsm=lockdown,yama,apparmor,bpf` to the `_custom_commandline` in your config, or add that to your bootloader's cmdline (which you can do on your current kernel without recompiling), or make a config fragment with the option turned on (see related option in customization.cfg).

jollycar:
thanks for the information. I got it working with _custom_commandline

PR #198, opened 2021-03-09
Add GCC version future version support

aerayth:
Fedora 34 and Rawhide both use GCC version 11, I have simply added a GCC version check which allows for older and newer versions to be checked with the CPU optimizations patch.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -250,7 +250,7 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   esac
 
   if [ "$opt_alternative_url" != "true" ]; then
-    wget "https://raw.githubusercontent.com/graysky2/kernel_gcc_patch/master/more-uarches-for-gcc-v10-and-kernel-${opt_ver}.patch"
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
That doesn't work on distros using GCC 10.2.0 for example.

What's wrong with using the native option in the meantime though?

aerayth:
I just modified the install.sh for myself, you could simply do an if statement check possibly to check if v11 is being used, I am not familiar enough with it to do checks for versions other than 10-11, etc.

Fedora 34/Rawhide come with v11 preinstalled, so you have to modify this to compile the kernel in my cases if you use GCC compiler option in the install.sh.

Tk-Glitch:
My problem with GCC 11 is that it's not released yet. Using the native option should work fine with GCC 11 still if you want to use the Zen3 optimizations. Also in case it was unclear, there is no change besides added Zen3 support between the GCC 10 and GCC 11 patch. You can use the GCC 10 patch on GCC 11 just fine.

Tk-Glitch:
I assume you were under the impression that the patch wouldn't work with GCC 11?

PR #202, opened 2021-03-15
Fix _cpusched for RC kernels

flgx16:


Tk-Glitch:
Goot. Thanks! :frog: :heart:

PR #206, opened 2021-03-22
Add mm-soft-dirty patchset from Paul Gofman

Joshua-Ashton:
Needed for D3D12 APITracing under Wine


Tk-Glitch:
Thanks Josh! :frog: :heart:

PR #207, opened 2021-03-22
Fix compilation on Ubuntu due canonical-certs

iWeaker4you:
Apparently in the new versions of the kernel or some configuration that I do not know but with this it is solved, almost very similar to Debian.


Tk-Glitch:
Ok then. Thanks! :frog: :heart:

PR #209, opened 2021-03-26
Some extra features for convenience 

### LINKED ISSUE #173: ./install.sh script fails on Fedora 33 KDE
I am on Fedora 33 and wanted to compile your kernel. with ./install.sh install script. It looks like the script is giving me error. What did I do before running the script? I configured the customization file, set the  linux distribution part to fedora, then set config expert to true. What kind of error did I received?  
This https://privatebin.net/?dc77bec5d9341a34#FwK9LJRANaUaMiZkjQ7CYXzCwP9hNRerPrhdgs77jD4A   

 Did I try to run the script with root privileges in another try? Yes I did. Still failed.(edited)
What other files should I provide you in order to fix this?
I wanted to compile UPDS with skylake patches.
but failed on the header compile part.

### LINKED ISSUE #204: CCache breaks install on OpenSUSE
Hi, I think I found an issue in the linux-tkg build. Whenever ccache is enabled on OpenSUSE Leap, the build fails with this error: https://pastebin.com/5BMbPRVD
This is on a fresh install of the OpenSUSE server edition

Disabling ccache in customization.cfg fixes this.

AdelKS:
Hello!

Some extra features that I personally find handy:
- Have the script copy the generated `.config` file next to the PKGBUILD, as `.config.new` so I can save it somewhere, after I did some various changes with `make xconfig`. I find the final `.config` file more useful than frags because the originally used `.config` file is not constant in non arch. 
- With the previous feature, the next step is to enable `install.sh` to use such custom `.config` files.
- Fixed an issue with `install.sh` where the prompt for the kernel version doesn't do newlines
- Enable `setversion` for `install.sh`, it may fix #173 and fix #204 actually.
- Enable using user provided `modprobed.db` databases, to compile stripped down kernels for friends.

What do you think of all this ?

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -44,6 +44,9 @@ _noccache="false"
 # !!!! Make sure to have a well populated db !!!! - Leave empty to be asked about it at build time
 _modprobeddb="false"
 
+# modprobed-db database file location
+_modprobeddb_db_path=~/.config/modprobed.db
+
 # Set to "1" to call make menuconfig, "2" to call make nconfig, "3" to call make xconfig, before building the kernel. Set to false to disable and skip the prompt.
 _menunconfig=""
 
--- a/customization.cfg
+++ b/customization.cfg
@@ -58,9 +58,16 @@ _diffconfig_name=""
 
 #### KERNEL OPTIONS ####
 
-# [Arch specific] Name of the default config file to use from the linux???-tkg-config folder. 
-# Default is "config.x86_64" and Arch hardened is "config_hardened.x86_64".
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Fine by me!

AdelKS:
I would like to rename `.config.new` to `kernelconfig.new` so it's not hidden by default, it bothered me a bit to need to "show hidden files" to see it. Then wer'e good for merger!

AdelKS:
Alright! We are all set here too!

One merge please -frog v2
@Tk-Glitch

Tk-Glitch:
Thanks Adel! Let's frog this up :frog:

PR #210, opened 2021-03-26
Add toggle for default TCP algorithm

### LINKED ISSUE #160: Switch from YeAH to BBR
BBR seems like the better choice, especially when you have a look at the round trip (ping) time. Here some numbers:

Algorithm | Average goodput [Mbps] | Average goodput ratio [%] | Average cwnd [#packets] | Average RTT [ms] | Average sending rate [Mbps] | Average throughput [Mbps]
----------|------------------------|---------------------------|-------------------------|------------------|-----------------------------|--------------------------
YeAH | 68.501 | 94.82 | 339.77 | 54.73 | 72.243 | 71.636
BBR | 67.442 | 95.28 | 31.22 | 4.37 | 70.779 | 70.527

### LINKED ISSUE #172: yeah is not selected by default by 0003-glitched-base.patch
applying patch `0003-glitched-base.patch` will not set yeah as default congestion algorithm, by itself, because default arch config is setting `CONFIG_TCP_CONG_CUBIC=y` and `CONFIG_TCP_CONG_YEAH=m`.

Must be `CONFIG_TCP_CONG_YEAH=y` to select in config as default. 
It can be verified with `cat /proc/sys/net/ipv4/tcp_congestion_control`, or by examine generated config that cubic is default with patch or without

AdelKS:
closes #172 and closes #160

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -178,6 +178,12 @@ _aggressive_ondemand="true"
 # https://github.com/Tk-Glitch/PKGBUILDS/issues/263
 _disable_acpi_cpufreq=""
 
+# [Advanced] Default TCP IPv4 algorithm to use. Options are: "yeah", "bbr", "cubic", "reno", "westwood". Leave empty if unsure.
+# This config option will not be prompted
+# Can be changed at runtime with the command line `# echo "$name" > /proc/sys/net/ipv4/tcp_congestion_control` where $name is one of the options above.
+# Default (empty) and fallback : cubic
+_tcp_cong_alg=""
+
 # You can pass a default set of kernel command line options here - example: "intel_pstate=passive nowatchdog amdgpu.ppfeaturemask=0xfffd7fff mitigations=off"
 _custom_commandline="intel_pstate=passive"
 
--- a/customization.cfg
+++ b/customization.cfg
@@ -178,7 +178,7 @@ _aggressive_ondemand="true"
 # https://github.com/Tk-Glitch/PKGBUILDS/issues/263
 _disable_acpi_cpufreq=""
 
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
I'm not sure we want it as default though :frog: Maybe worth another toggle?

AdelKS:
Oh! I thought you wanted it as default since it's written in the README. I do not have a strong opinion to be honest otherwise x)

AdelKS:
Two possibilities:
- You can actually just update the README and remove that line, then we can close this PR without merger. Speaking of, I need to update my README PR to fit with the recent changes.
- I can make this into a toggle for default algo: Cubic, yeah, bbr and maybe some others

Tk-Glitch:
It can't be set to default "as is" due to missing bits in the kernel. We have a patch that simply allows one to set it as the default if desired, but indeed it requires the user to enable it through config.

Tk-Glitch:
A config option is fine by me. I don't think we want a prompt for that, just the config option would be better imho.

AdelKS:
Good! And we keep Cubic as default and fallback. That would close #172 and #160. Two less issues open! Gotta love to reduce that number.

AdelKS:
Alright! Tests have been done, it works as intended.

One merge please -frog v2
@Tk-Glitch

AdelKS:
Oh shit wait, I just saw a "k10" tcp algorithm... I frogged up xD

Lemme fix that fast!

AdelKS:
Okay done. No one saw that copy-pasta error.

Tk-Glitch:
Thanks Adel! :frog: :heart:

PR #211, opened 2021-03-26
Fix urls for graysky2's cpu opt patchsets

bcvandendool:
The urls for graysky2's cpu opt patchests have been changed, see [this commit](https://github.com/graysky2/kernel_gcc_patch/commit/5fa7fc7802c929eb239198adabcfc42ee852065b), causing the script to fail on downloading them, this updates them to the new versions.


Tk-Glitch:
I've lost track of how many times those got renamed and moved in the last month. Thank you! :frog: :heart:

PR #217, opened 2021-04-05
 Fix Debian / Ubuntu compile due to .config

iWeaker4you:
Since the moment in which .config is created was changed, I transferred the Debian and Ubuntu workarounds to Prepare, it did not compile because it could not find the .conf when doing "sed"

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -260,16 +260,7 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   # cd in linux folder, copy Ubuntu's current config file, update with new params
   cd "$_where"/linux-src-git
 
-  if [ "$_distro" = "Debian" ]; then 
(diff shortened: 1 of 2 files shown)


Kerrung:
I have the same problem on Arch Linux. Doesn't compile because it can't find .config

iWeaker4you:
> I have the same problem on Arch Linux. Doesn't compile because it can't find .config

with Makepkg -si or ./install.sh install ?

Kerrung:
> > I have the same problem on Arch Linux. Doesn't compile because it can't find .config
> 
> with Makepkg -si or ./install.sh install ? 

$ makepkg -sri

Tk-Glitch:
Thanks! :frog: :heart:

PR #218, opened 2021-04-06
Arch: Fix failure to extract kernel patch when srcdest is set in makepkg.conf

Spoons:
On my machine, the kernel fails to build when srcdest is set in makepkg. My solution was to check the source folder and extract the upstream patch if the extracted patch is missing.

I am open to any changes.


Tk-Glitch:
Works for me. Thanks! :frog: :heart:

PR #220, opened 2021-04-13
Void no longer uses LibreSSL, switched to OpenSSL

Hyper-KVM:
No need to change checksumming to SHA1 anymore :)


Tk-Glitch:
Thank you! :frog: :heart:

PR #221, opened 2021-04-13
enable Clang LTO usage

Hyper-KVM:
This PR just allows use of LTO when compiling with clang if the user wants to.

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -84,6 +84,11 @@ _cpusched=""
 # For advanced users.
 _compiler=""
 
+# Clang LTO mode - options are "full" or "thin".
+# full: uses 1 thread for Linking, slow and uses more memory in exchange for faster runtime performance.
+# thin: uses multiple threads, faster and uses less memory in exchange for slower runtime performance.
+_lto_mode=""
+
 # CPU sched_yield_type - Choose what sort of yield sched_yield will perform
 # For PDS and MuQSS: 0: No yield. (Recommended option for gaming on PDS and MuQSS)
 #                    1: Yield only to better priority/deadline tasks. (Default - can be unstable with PDS on some platforms)
--- a/customization.cfg
+++ b/customization.cfg
@@ -85,6 +85,7 @@ _cpusched=""
 _compiler=""
 
 # Clang LTO mode - options are "full" or "thin".
+# Linux kernel >=5.12 Only.
 # full: uses 1 thread for Linking, slow and uses more memory in exchange for faster runtime performance.
 # thin: uses multiple threads, faster and uses less memory in exchange for slower runtime performance.
 _lto_mode=""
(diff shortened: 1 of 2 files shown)


AdelKS:
I can't wait for the 5.12 release ! I think we should state that it's only available for 5.12 and onwards in the customization file.

linux-tkg-config/prepare
```diff
@@ -640,6 +649,14 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
       sed -i -e 's/CONFIG_GCC_PLUGINS=y/# CONFIG_GCC_PLUGINS is not set/' ./.config
     fi
   fi
+  # Clang LTO
+  if [ "$_compiler_name" = "-llvm" ] && [ $_basever -ge 512 ]; then
+    if [ "$_lto_mode" = "full" ]; then
+      echo "CONFIG_LTO_CLANG_FULL=y" >> ./.config
```
AdelKS, line 655: I think from now on it's better to use:
```
scripts/config -e LTO_CLANG_FULL
scripts/config -d LTO_CLANG_THIN
```
Instead of `echo` or `sed`. Since these options can already be in the config file (and that will happen for `install.sh` installs in the future), `scripts/config` knows when to add or replace existing lines in `.config`.

linux-tkg-config/prepare
```diff
@@ -640,6 +649,14 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
...
+    elif [ "$_lto_mode" = "thin" ]; then
+      echo "CONFIG_LTO_CLANG_THIN=y" >> ./.config
```
AdelKS, line 657: Same applies here

linux-tkg-config/prepare
```diff
@@ -338,6 +343,10 @@ _tkg_srcprep() {
 
   msg2 "Setting version..."
   scripts/setlocalversion --save-scmversion
+  if [ $_basever -ge 512 ] && [[ "$lto_enable" =~ [yY] ]] || [ -n "$_lto_mode" ]; then
+    scripts/config -e LTO_CLANG_THIN
```
AdelKS, line 347: Better be safe here and explicitly disable full lto with `scripts/config -d LTO_CLANG_FULL`, since a user may have a full LTO config from the currently running kernel and wants to switch to thin LTO.

customization.cfg
```diff
@@ -84,6 +84,12 @@ _cpusched=""
 # For advanced users.
 _compiler=""
 
+# Clang LTO mode - options are "full" or "thin".
```
AdelKS, line 87: We should also explicitly state that it's only available when choosing `llvm` as a compiler. Some of us don't know that `llvm` or `clang` in our case it's the same thing.

PR #222, opened 2021-04-14
install.sh: improve git fetching and stop producing source code packages

AdelKS:
Hello!

Here I am for another round of improvements on `install.sh`. On the menu:
- Now compiling `bindeb-pkg` and `binrpm-pkg` instead of `deb-pkg` and `rpm-pkg`. When I first implemented it I thought the source code ones are needed for DKMS things. But afaik only the headers are needed. It makes the packaging process, after compilation, way faster.
- Git shallow remote fetching wasn't optimal. Fixed that
- Handling two remotes in Git wasn't done properly, fixed that.
- A few uses of reggex'es to makes conditions smaller.

I tested my changes (except on Suse, I got lazy) and I think it's all good to go!

One merge please -frog v2

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -175,11 +175,11 @@ srcdir="$_where"
 
 source customization.cfg
 
-source linux-tkg-config/prepare 
+source linux-tkg-config/prepare
 
 if [ "$1" != "install" ] && [ "$1" != "config" ] && [ "$1" != "uninstall-help" ]; then
   msg2 "Argument not recognised, options are:
-        - config : interactive script that shallow clones the linux 5.x.y git tree into the folder linux-src-git, then applies extra patches and prepares the .config file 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -303,16 +303,16 @@ if [ "$1" = "install" ]; then
 
   # ccache
   if [ "$_noccache" != "true" ]; then
-
-    if [ "$_distro" = "Ubuntu" ] || [ "$_distro" = "Debian" ]; then
+    # Todo: deal with generic and paths, maybe just export boths possibilities and not care
```
AdelKS, line 306: Oh this line I forgot to strip it. It's from my other work on making a generic install work (`make install` and co). Since my commits are super clean. Let's just leave it here and say no one ever saw this useless comment.

Tk-Glitch: :frog:

Tk-Glitch:
Let's go. Thanks Adel! :frog: :heart:

PR #227, opened 2021-04-18
install.sh: Add convenience script for generic install

AdelKS:
Hello,

This is  something I personally use on my Gentoo, which is basically the generic way to install the kernel:
```shell
sudo make modules_install
sudo make install
```
I added two more commands since it's what I personally use:
```shell
sudo dracut --hostonly --kver $_kernelname
sudo grub-mkconfig -o /boot/grub/grub.cfg
```
I don't know if it can be useful for others. What do you think ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,3 +119,11 @@ The command to do for that is:
 ./install.sh config
 ```
 
+If one chooses `Generic` as distro. `./install.sh install` will compile the kernel then prompt before doing the following:
+```shell
+sudo make modules_install
+sudo make install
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Will need rebasing, otherwise fine by me.

AdelKS:
Looks like I forgot to force push my changes last time I worked on it xD
Here we go!

Tk-Glitch:
Thanks Adel! :frog: :heart;

PR #230, opened 2021-04-20
Build kernel-devel package for Fedora

trgeiger:
Fedora does require the kernel-devel package for DKMS, as you can see when running `rpm -q --requires dkms` to check for DKMS dependencies. It lists `kernel-devel-uname-r` as a dependency.

This commit just reverts some of the recent changes for Fedora systems so that kernel-devel is created by the install script.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -180,7 +180,7 @@ source linux-tkg-config/prepare
 if [ "$1" != "install" ] && [ "$1" != "config" ] && [ "$1" != "uninstall-help" ]; then
   msg2 "Argument not recognised, options are:
         - config : interactive script that shallow clones the linux 5.x.y git tree into the folder linux-src-git, then applies extra patches and prepares the .config file
-                   by copying the one from the currently running linux system and updates it. 
+                   by copying the one from the currently running linux system and updates it.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -392,10 +392,11 @@ if [ "$1" = "install" ]; then
         fi
         _headers_rpm="kernel-headers-${_kernelname}*.rpm"
         _kernel_rpm="kernel-${_kernelname}*.rpm"
+        _kernel_devel_rpm="kernel-${_kernelname}*.rpm"
```
AdelKS, line 395: Looks like a copy-pasta error here :yum:

AdelKS:
If it is fine with you @trgeiger, I will open a PR to revert my changes for Fedora and Suse, but not entirely since now I know that the unneeded package is actually `kernel-headers` and not `kernel-devel`. Which is weird since the headers are actually contained in `kernel-devel` and not `kernel-HEADERS`, but whatever.

install.sh
trgeiger: oh whoopsie. your PR looks better overall anyway ;)

AdelKS: Suse would have had the same issue so it needed to be handled too, thanks for reporting!

Issue #229, opened 2021-04-20
kernel-devel actually is needed on Fedora for DKMS


trgeiger:
The recent change to no longer build kernel-devel for Fedora has broken DKMS. Seems that Fedora requires the corresponding -devel package for the current kernel to install DKMS modules.
```bash
❯ sudo dkms autoinstall          
Error! echo
Your kernel headers for kernel 5.11.15_tkg_pds cannot be found at
/lib/modules/5.11.15_tkg_pds/build or /lib/modules/5.11.15_tkg_pds/source.
You can use the --kernelsourcedir option to tell DKMS where it's located.
```

trgeiger:
I opened this PR to address this https://github.com/Frogging-Family/linux-tkg/pull/230

AdelKS:
I expected the headers to be installed by the... `kernel-headers` RPM package x)

The folder `/lib/modules/5.11.15_tkg_pds/build` seems to be pointing to `/usr/src/kernels/5.11.15_tkg_pds` which for sure is installed by the `kernel-devel` package. `kernel-headers` seems to install stuff in `/usr/include` and actually this is he useless package that can be omitted :thinking:

AdelKS:
Okay, I just checked what's inside the DEB packages and they got their naming right : `linux-headers` actually installs the headers that DKMS uses. `linux-libc-dev` DEB package is the equivalent of `kernel-headers` in Fedora, confusing stuff. It really seems like Fedora packaging script on the kernel side has an issue...

PR #233, opened 2021-04-20
Fix DKMS for RPM distros

AdelKS:
Weirdly enough, I expected the kernel headers to be contained in the `kernel-headers` RPM packages. But they are actually in `kernel-devel`. So obviously that broke DKMS, and now it should be fixed!. Although we are back to using the lenghty `make rpm-pkg`.

I checked what's inside the DEBs and my change should work since `linux-headers` DEB package does actually contain the headers where DKMS wants them.

Closes: #230
Closes: #229

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -339,7 +339,7 @@ if [ "$1" = "install" ]; then
 
       # Create DEBS folder if it doesn't exist
       mkdir -p DEBS
-      
+
       # Move rpm files to RPMS folder inside the linux-tkg folder
       mv "$_where"/*.deb "$_where"/DEBS/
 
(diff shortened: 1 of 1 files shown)


trgeiger:
Only thing I'm not sure about is if SUSE handles headers the same way as Fedora

AdelKS:
I expect the RPM creation process to be independent from the distro since the scripts are contained in the kernel sources, one can theoretically make RPMs in other distros too, although there are distro dependent RPM packaging packages that are needed  :thinking: 

But fair point, I will fire up a tumbleweed VM to check I guess. Or  wait for a bug report :laughing:

AdelKS:
I gave it a try too, the `kerne-devel` package is indeed the same in tumbleweed. All good!

Tk-Glitch:
Interesting issue. Thanks guys!

PR #239, opened 2021-05-02
Fix: make Arch be able to use _configfile

AdelKS:
Fix: make Arch be able to use `_configfile` because the wrong path was being used. I fixed also a wrong use of `_msg2` I think.

Also, add `yes ""` to all the `make` calls, sometimes, I get prompts.

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -42,6 +42,7 @@ _force_all_threads="true"
 _noccache="false"
 
 # Set to true to use modprobed db to clean config from unneeded modules. Speeds up compilation considerably. Requires root - https://wiki.archlinux.org/index.php/Modprobed-db
+# Using this option can trigger user prompts if the config doesn't go smoothly.
 # !!!! Make sure to have a well populated db !!!! - Leave empty to be asked about it at build time
 _modprobeddb="false"
 
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -344,9 +344,9 @@ _tkg_srcprep() {
 
   msg2 "Setting version..."
   scripts/setlocalversion --save-scmversion
-  
+
   if [ "${_distro}" = "Arch" ]; then
-    
+
     echo "-$pkgrel-tkg-${_cpusched}${_compiler_name}" > localversion.10-pkgrel
     echo -e "Version tail set to \"-$pkgrel-tkg-${_cpusched}${_compiler_name}\"\n" > "$_where"/prepare.log
     echo "" > localversion.20-pkgname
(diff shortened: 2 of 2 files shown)


linux-tkg-config/prepare
```diff
@@ -506,23 +506,24 @@ _tkg_srcprep() {
...
+        msg2 "Using /proc/config.gz as config file"
         zcat --verbose /proc/config.gz > .config
       else
-        _msg="Current kernel config not found! Falling back to default..."
+        msg2 "Current kernel config not found! Falling back to default..."
       fi
     fi
   else
-    cat "${srcdir}/${_configfile}" > ./.config
+    msg2 "Using user-provided config file in ${_where}/${_configfile}"
```
AdelKS, line 525: Okay, here if one uses `config_hardened.x86_64` it will still say this sentence. Maybe I should make a condition ?

Tk-Glitch:
After looking at the lost HZ values last time, I noticed that we can end up with wrong values for stuff that isn't just a boolean y/n. So I'd prefer those to go, and we'll fix/silence the prompts by hand as needed.

AdelKS:
You are talking about the `yes "" |` that you prefer to not be there ? If so I will revert that commit.

About the HZ configs, I will PR later on a more elegant approach with the `scripts/config` script just like we do with the TCP algs if you are fine with that.

Tk-Glitch:
Yes indeed. `yes "" |` hid the issue that was apparently there since it's inclusion, so I'm not a huge fan of using it moving forward. Regarding HZ config handling, I'm fine with moving to `script/config` for it.

AdelKS:
I understand, and actually enabling the use of `_configfile` will make everything more complicated if we stick to using `sed`s to modify it: initially you were using `sed` because you knew exactly what's in the original `.config` file, which is arch's official one. Now we won't know any longer and I think all the `sed`s should be replaced with `scripts/config` where possible. I have the feeling that the HZ issue is something related to this.

I will remove all the `yes "" |` and we'll make sure we edit stuff in a way that `make` won't prompt stuff, and if it does, then we would need to work it out.

Tk-Glitch:
The HZ issue happened on Manjaro, so using the Arch path. The thing was that it was supposed to prompt (without `yes "" |`) due to an additional option, and didn't, defaulting to a wrong value and overwriting the chosen one.
For a few things, sed might/will still be needed, I think. But we can probably nuke most of them yeah.

AdelKS:
Okay I removed that one use of `yes` from `make localmodconfig` call and added info in `customization.cfg` to tell it can prompt.

linux-tkg-config/prepare
```diff
@@ -506,23 +506,24 @@ _tkg_srcprep() {
...
-  
-  if [ -z "${_configfile}" ]; then    
+
+  if [ -z "${_configfile}" ]; then
     if [ "${_distro}" = "Arch" ] || [ "$_distro" = "Void" ]; then
       cat "${srcdir}"/config.x86_64 > ./.config
     else
       if [ -f /boot/config-`uname -r` ];then
-        _msg="Using /boot/config-`uname -r` as config file"
+        msg2 "Using /boot/config-`uname -r` as config file"
```
Tk-Glitch, line 515: Please respect the switch to the _msg variable for patch application messages :frog:

AdelKS: That's the thing, it's not a patch, just an informative message :thinking:

Tk-Glitch: It was changed to allow for showing the messages dynamically:
https://github.com/Frogging-Family/linux-tkg/commit/09b5e06b5cee80a9a25ef71cd64ed2774cd14009

customization.cfg
```diff
@@ -42,6 +42,7 @@ _force_all_threads="true"
 _noccache="false"
 
 # Set to true to use modprobed db to clean config from unneeded modules. Speeds up compilation considerably. Requires root - https://wiki.archlinux.org/index.php/Modprobed-db
+# Using this option can trigger user prompts of the config doesn't go smoothly.
```
Tk-Glitch, line 45: > prompts of the config

I think you meant `if` instead of `of` here.

linux-tkg-config/prepare
Tk-Glitch: Oh I see what you mean. I should go to sleep.

AdelKS: Yes, and as I understand things `_tkg_patcher` needs to be called for the message to be displayed. It's not called here and `msg2` gets overridden after. For sure I don't understand something, I will look deeper into it.

customization.cfg
AdelKS: Absolutely x)

AdelKS: I will changed to commit message with the double "t" too

Tk-Glitch:
Thanks Adel! :frog: :love:

PR #243, opened 2021-05-16
Fix winesync for non-arch distros

AdelKS:
Hello,

I was kinda expecting things to Just Work™ with the winesync changes. But it was not the case, it was simply not working for non-arch:

- The header in `/usr/include/linux/winesync.h` was not installed, thanks (#sarcasm) to my previous change to make things faster for DEB and RPM distros that already got partly reverted. Now it is fully reverted for them. And I added installing those files for the Generic approach
- Install the `winesync.rules` file "manually". For now that file can only be cleaned up manually. I do not think it's a big deal: it's only used by us, and if we need to change it I will add an update step to the `install.sh` script.
- Clean up the `winesync.rules` file on exit.

That's about it. Maybe I still miss something to make things work properly between `wine-tkg` and `linux-tkg` for non-arch, feedback is most welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,10 @@ The command to do for that is:
 If one chooses `Generic` as distro. `./install.sh install` will compile the kernel then prompt before doing the following:
 ```shell
 sudo make modules_install
+sudo make headers_install INSTALL_HDR_PATH=/usr # CAUTION: this will replace files in /usr/include
 sudo make install
-sudo dracut --hostonly --kver $_kernelname
+sudo dracut --force --hostonly --kver $_kernelname
 sudo grub-mkconfig -o /boot/grub/grub.cfg
 ```
+**Note:** these changes will not be tracked by your package manager and uninstalling requires manual intervention.
 
--- a/install.sh
+++ b/install.sh
@@ -292,6 +292,12 @@ fi
 
 if [ "$1" = "install" ]; then
 
+  if [ -e "${_where}/winesync.rules" ]; then
+    msg2 "Installing udev rule for winesync"
+    sudo cp "${_where}"/winesync.rules /etc/udev/rules.d/winesync.rules
+    sudo chmod 644 /etc/udev/rules.d/winesync.rules
+  fi
+
   # Use custom compiler paths if defined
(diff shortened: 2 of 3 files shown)


Tk-Glitch:
What about https://github.com/Frogging-Family/linux-tkg/blob/master/PKGBUILD#L613 ? :frog:

AdelKS:
> What about https://github.com/Frogging-Family/linux-tkg/blob/master/PKGBUILD#L613 ? frog

I thoroughly checked and the header is actually picked up by the packaging script for DEB and RPM distros with their third package that I stopped installing because it wasn't useful for DKMS, but it is useful for `winesync` since it installs that header.  `make headers_install` installs it in the Generic approach. There is only one version of `winesync.h` right ?

Tk-Glitch:
There's only one version, but wine won't pick it up in its default location in the kernel headers, which is why it's being copied over.

AdelKS:
Wait, it needs to be here right: `/usr/include/linux/winesync.h` ? Because that's where it is installed by default already :thinking:

Tk-Glitch:
Well, if that's where it ends up on non-Arch by default I guess we're good to go.

AdelKS:
Wait, I only checked `Generic` and `Ubuntu`. I just checked Fedora: we're all good!

To summarize:
- Have the module built, installed, and auto-started. Oh right I did not install anything in `/etc/modules.load.d` I should add that.
- Have the udev rule set
- Have the header at the correct location, `wine-tkg` can only be properly compiled once this header is there right ? If so, it seems that it can compile even without that header there :thinking: maybe do a check in `wine-tkg`'s script.

Am I missing anything else ?

Tk-Glitch:
Yeah, the patch used to break when the header was missing, but Zeb changed it so it doesn't. I might add a check.

Looks fine to me otherwise.

AdelKS:
Then we are good to go for now. Last time I tried in my `Generic` distro approach I couldn't get `fastsync` to properly work. I haven't gotten to the bottom of it yet. You can merge this, and if anything else needs to be change I will PR again, since you have already reviewed these changes.

Tk-Glitch:
LET'S-A GO! :frog:

PR #248, opened 2021-05-22
Various refactorings + Enable Clang LTO

AdelKS:
Hello!

I wanted to finish @Hyper-KVM's work and I ended up refactoring few stuff :+1: 
- Removed the use of the files `cpuschedset`, `compilerset` and `versel`
- Discovered the existence of `select` and used it for other stuff. I also used associative arrays to implement prompting without redundant lines of code
- Moved the code paths that set the compiler, the cpu scheduler and the kernel version into separate functions
- LTO is properly handled although it does not compile for me.

This should close #221


AdelKS:
- `select` is not longer  used as I implemented a `_prompt_from_array` function that can take into account a default entry that is return if the user presses return without text input.
- `kernel` sub-versions are to be edited directly in an associative-array. The use of the associative array simplifies the code that prompts and sets the kernel version.
- moved nearly all the `sed`s that edit the `.config` file to the use of `scripts/config`, except the Futex2 enablement configs because there's no use to do that.

linux-tkg-config/prepare
```diff
@@ -797,68 +746,81 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
...
       _sched="pds"
-    elif [ "${_cpusched}" = "MuQSS" ] || ( [ "$_basever" != "54" ] && [ "$_cpusched" != "bmq" ] && [ "$_cpusched" != "pds" ] ) || ( [ "$_basever" = "54" ] && [ "$_cpusched" = "bmq" ] ); then
+    elif [ "${_cpusched}" = "muqss" ] || ( [ "$_basever" != "54" ] && [ "$_cpusched" != "bmq" ] && [ "$_cpusched" != "pds" ] ) || ( [ "$_basever" = "54" ] && [ "$_cpusched" = "bmq" ] ); then
       _sched="${_cpusched}"
     else
       _sched="alt_core"
     fi
-    if [ "$CONDITION0" = "0" ]; then
+    if [ "$_sched_yield_type" = "0" ]; then
       sed -i -e 's/int sched_yield_type __read_mostly = 1;/int sched_yield_type __read_mostly = 0;/' ./kernel/sched/${_sched}.c
```
Tk-Glitch, line 817: For muqss, the file name is `MuQSS.c`. I don't see it handled here, which would lead to `muqss.c` -> not found. Did I miss it?

AdelKS: No you did not, I just didn't notice this small thing that makes a huge difference. Will fix this

AdelKS: It should be good now.

AdelKS: No nevermind, I still need to fix few things

linux-tkg-config/prepare
```diff
@@ -1136,48 +1028,31 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
...
+    _tickless_array_text=(
+      "No, use periodic ticks."
+      "Yes, full tickless baby!\n        Can give higher performances in many cases but lower consistency on some hardware."
+      "Just tickless idle plz.\n        Just tickless idle can perform better with some platforms (mostly AMD) or CPU schedulers (mostly MuQSS)."
+    )
+    _default_index="2"
+    _prompt_from_array "${_tickless_array_text[@]}"
+    _tickless="${_selected_index}"
+  fi
+  if [ "$_tickless" = "2" ]; then
```
Tk-Glitch, line 1047: This  should be 0

Tk-Glitch:
`_undefine()` nukes entries from the config, breaking unattended building. We'd want to preserve the unset options.

Tk-Glitch:
Looking at the script/config code, we'd want to use --disable for such cases:
```
	--disable|-d)
		set_var "${CONFIG_}$ARG" "# ${CONFIG_}$ARG is not set"
		;;
```
```
	--undefine|-u)
		undef_var "${CONFIG_}$ARG"
		;;
```

linux-tkg-config/prepare
```diff
@@ -662,457 +620,392 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
...
     fi
   elif [ "$_compileroptlevel" = "2" ]; then
-    sed -i -e 's/CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE=y/# CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE is not set/' ./.config
+    _undefine "CC_OPTIMIZE_FOR_PERFORMANCE"
     if [ "$_basever" != "54" ]; then
-      echo "CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE_O3=y" >> ./.config
+      _enable "CC_OPTIMIZE_FOR_PERFORMANCE_O3"
     else
-      echo "CONFIG_CC_OPTIMIZE_HARDER=y" >> ./.config
+      _enable _undefine "CC_OPTIMIZE_HARDER"
```
Tk-Glitch, line 912: That _undefine looks like a typo

AdelKS:
Oh I thought `undefine` does the "is not set" thing whereas `disable` puts it equal to "n". But now that I think of it, the later doesn't exist... xD will fix that.

linux-tkg-config/prepare
AdelKS: Absolutely

linux-tkg-config/prepare
```diff
@@ -662,457 +620,392 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\n
...
+    if [ "$_processor_opt" = "$_march" ]; then
+      if [[ "$_march" == "generic"* ]]; then
+        _enable "${_march}"
       else
-        echo "# CONFIG_M${_march_upper} is not set" >> ./.config
+        _enable "M${_march}"
       fi
-    done
-  fi
+    else
```
Tk-Glitch, line 721: You'll also need to handle the generic marchs in the disable side, else we're getting prompted even though we have selected the march we wanted already:
```
  for _march in "${_cpu_marchs[@]}"
  do
    if [ "$_processor_opt" = "$_march" ]; then
      if [[ "$_march" = "generic"* ]]; then
        _enable "${_march}"
      else
        _enable "M${_march}"
      fi
    else
      if [[ "$_march" = "generic"* ]]; then
        _disable "${_march}"
      else
        _disable "M${_march}"
      fi
    fi
  done
```
(taking the _undefine -> _disable change into account)

Tk-Glitch:
Another problem is that we were using the cpuschedset file as a state tracker to prevent makepkg from looping the prompt(s). With your changes, it'll currently prompt again about version, cpu scheduler and compiler (and maybe also profiles) before packaging, which is undesirable.

Tk-Glitch:
So yeah, looks like we'll need to have a file containing basic dynamic settings. This works:

PKGBUILD
Adding a check around _tkg_initscript :
```
   source "$_EXT_CONFIG_PATH"
 fi
 
+# Make sure we're in a clean state
+if [ ! -e "$_where"/BIG_UGLY_FROGMINER ]; then
+  _tkg_initscript
+fi
+
 if [[ "$_sub" = rc* ]]; then
   _srcpath="linux-${_basekernel}-${_sub}"
   kernel_site="https://git.kernel.org/torvalds/t/linux-${_basekernel}-${_sub}.tar.gz"
```

prepare
Sourcing the frogminer early with failure being considered as clean state :
```
   ["5.13"]="rc4"
 )
 
+ source "$_where"/BIG_UGLY_FROGMINER >/dev/null 2>&1 || msg2 "No frogminer found, assuming clean state.."
+ 
 _undefine() {
   for _config_name in "$@"; do
     scripts/config --undefine "${_config_name}"
```

Storing _set_kernel_version() variables in frogminer:
```
   _basever=`echo $_version | tr -d "."`
   _sub=${_kver_subver_map[$_version]}
 
+  echo -e "_basekernel='$_basekernel'\n_basever='$_basever'\n_sub='$_sub'" > "$_where"/BIG_UGLY_FROGMINER
 }
 
 _set_cpu_scheduler() {
```

Storing _set_cpu_scheduler() variables in frogminer:
```
     _cpusched=${_avail_cpu_scheds[$_selected_index]}
   fi
 
+  echo -e "_cpusched='$_cpusched'" >> "$_where"/BIG_UGLY_FROGMINER
 }
 
 _set_compiler(){
```
Storing _set_compiler() variables in frogminer:
```
   fi
 
+  echo -e "_compiler_name='$_compiler_name'\nllvm_opt='$llvm_opt'" >> "$_where"/BIG_UGLY_FROGMINER
 }
 
 
```

Getting rid of the frogminer on exit:
```
   rm -f "$_where"/cleanup
   rm -f "$_where"/prepare
 
+  # Remove state tracker
+  rm -f "$_where"/BIG_UGLY_FROGMINER
+
   # Remove winesync rules file
   rm -f "$_where"/winesync.rules
 
```

linux-tkg-config/prepare
AdelKS: Yes I forgot about it x) Thanks!

AdelKS:
Yea I asked why we were using files to store variables, which looked like double work to me, on Discord but I didn't insist on you answering x) Will implement those changes tonight !

Tk-Glitch:
> Yea I asked why we were using files to store variables, which looked like double work to me, on Discord but I didn't insist on you answering x) Will implement those changes tonight !

Sorry about that, I probably got distracted.

Tk-Glitch:
There are other issues. We're losing some naming parts. Trying to find where.

linux-tkg-config/prepare
```diff
@@ -554,29 +515,26 @@ _tkg_srcprep() {
     cat "${_where}/${_configfile}" > ./.config
   fi
 
+  if [ "${_distro}" = "Arch" ]; then
+    # Reset local version string if ever it's in the .config file
+    scripts/config --set-str localversion ""
```
Tk-Glitch, line 520: I still fail to see the point of this when we are running `make LOCALVERSION=` in the PKGBUILD. Were your tests on Arch done with the PKGBUILD or with install.sh regarding that versioning duplication problem?

AdelKS: I did it on Arch, I think it is reproducible if you have a config file where `LOCALVERSION` is explicitly set or you set it with `make xconfig`.

AdelKS:
> There are other issues. We're losing some naming parts. Trying to find where.
edit: Found it and fixed in the code I posted above.

You can push the changes directly here if you wish :D  I have a few changes in mind that I still need to implement: 
- Use an associative array for Project C revisions
- What do you think of modularizing more the code  by moving chunks into functions, so `_tkg_initscript` and `_src_prepare` will just be a bunch of one liners, that would help you use the equivalent of the `_debug` variable by simply commenting the one-liners you want. Code chunks to move :
  - Fsync/Futex2 enablement
  - Graysky CPU march patching and selection
  - TCP algs
  - Scheduler selection, patching and enablement
  - misc
  - etc...

Tk-Glitch:
tbh all of this has never been a problem. None of these changes are *necessary*. The problem is maintenance. Right now adding or remove support for a kernel is annoying. Before, when we had a single dir for a given kernel, it was just a matter or deleting it and done in a sec. Now it requires some time (probably a minute or two) and focus (not removing a check for another version by mistake), which makes all of this a pain in the ass. It might look fancy but I don't even feel like I want to bother cleaning up the legacy kernels support. Also, removing it means people who'd want to compile an older -now unsupported- kernel will have to search through the git index to find what they want, instead of having a dir ready for grab in the legacy branch.

So don't overthink it, it's just a script to install kernels. There's no point in over-engineering it if it doesn't actually provide more than it already does.

Doing it for fun is fine, of course, but my last point still stands.

AdelKS:
I understand your point and your concern, 

I think the current script has quite a bit of redundancies, a little bit hard to read through, but does the job. I am doing this because I like contributing to `linux-tkg` and I like to think of elegant and short approaches (so, this is the fun part), but I do not think I am doing it at the expense of readability or over-engineering. Quite the opposite actually: adding `CacULE` was simple, making it the only one to support `2000Hz` was simple. Prompting with defaults is simple and less redudant... etc

I think the main benefit of my changes is the use of associative arrays, and they work towards actually making it easy to remove old kernel versions by just removing it from the arrays instead of going through each hand-written condition. It is also important to make the switch from `sed` to `scripts/config` to properly handle custom config files. But I agree that this last part was my idea and using `fragments` was good and worked properly for Arch.

I brought useless complexity to the table because by pushing the support for other distros. I do think it's for the better as more people can enjoy this and make the community grow bigger. I find it stupid to just support Arch although a big user base is elsewhere. Also, this can help them move to Arch since they get exposure to it (this what happened with me).

I understand that reviewing this is taking much precious time from you that you would rather be spending doing more important work. I have not asked you about these changes beforehand so you could refuse, either. Although you did agree that a big part of the `sed` calls can be changed to `scripts/config`.

If you wish, you can close this PR and not merge it. I won't hold any personal grudge as I understand how much time-constrained you are. And I will stop doing PRs that add in complexity.

Tk-Glitch:
I wasn't being critical about these proposed changes, nor your past or future contributions and I'm sorry if you felt it that way. I was just pointing out that the goal of those scripts isn't necessarily requiring too much of your time. We can make it extremely clean and streamlined for sure, but I don't think it's *that* horrible in its current state ( :frog: ) and the main motivation should be to push things further for new desirable features integration (for users and devs/ourselves alike) first. This PR is nice and I have nothing against it btw, and I have pushed the changes we talked about earlier since your gave me your green light :wink:

AdelKS:
I am happy you like these changes :smile:

> I was just pointing out that the goal of those scripts isn't necessarily requiring too much of your time

I think you are right about this, I took some time with this even though it absolutely is not necessary. I could've spent that time to on my own software to finally release a new version, but my heart is with `linux-tkg`, what can I do x)

> I don't think it's that horrible in its current state ( :frog: ) and the main motivation should be to push things further for new desirable features integration (for users and devs/ourselves alike) first

It is not horrible at all, I got the feeling that this work had to be done sooner or later, but sooner is better since the code is getting bigger by the day and following the "old" verbose approach when adding features doesn't help. So now we can push new features on a more clean base with a "newer" approach :D 

> I have pushed the changes we talked about earlier since your gave me your green light wink

I always PR with that green light on, so you can just directly push anything that would make us both win time :D

Tk-Glitch:
More issues.

The way scripts/config works is by adding enabled/disabled stuff at the end of the file. Then when the config get regen'd, it's cleaned up and in most cases, will use the last defined value. However, it appears that it's not always the case.

With this for example:
```
    _disable "FONT_TER16x32"
    _enable "FONT_AUTOSELECT"
```
The end result will be :
```
CONFIG_FONT_TER16x32=y
```
And `CONFIG_FONT_AUTOSELECT` gets nuked out of the file.

Using the previous way works:
```
  if [ "$_font_autoselect" != "false" ]; then
    sed -i -e 's/CONFIG_FONT_TER16x32=y/# CONFIG_FONT_TER16x32 is not set\nCONFIG_FONT_AUTOSELECT=y/' ./.config
  fi
```

What's scary now is that we need to know when/if the setting will stick or not.

AdelKS:
What, it doesn't look through the config file first ?! I expected more from that script...

A way to work around it is to always nuke the entry first with `_undefine`, then use `_enable` or `_disable`. This should be consistent. Well I expected the nuke command to look through the config file, will check.

Tk-Glitch:
Yeah that's what I tried next. Sadly it seems to be inconsistent. I'll experiment with it.

Tk-Glitch:
Actually it adds `# CONFIG_FONT_TER16X32 is not set` while the option is `CONFIG_FONT_TER16x32` (note the non-capital x) :frog:

AdelKS:
Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

AdelKS:
Right, it auto uppercases everything. Forgot about it, I suppose I expected the config options to never use lowercase x)

Tk-Glitch:
Looks like `--keep-case` is an option

AdelKS:
Yep absolutely!

AdelKS:
If we are going to use `--keep-case` all the time I need to change a few lowercase stuff that I was too lazy to write in capital letters

Tk-Glitch:
Any other that I might have missed? I think we're good.

AdelKS:
I am in front of my computer now !

So, finally the `scripts/config` works as as expected I suppose ?

Unfortunately the `${_march^^}` approach doesn't work with `zsh`, only `tr` seems like a universal thing last time I googled.

Tk-Glitch:
Yep it's finally working as expected.

> Unfortunately the ${_march^^} approach doesn't work with zsh

tr it is then!

Okay it seems to be fine now. I'm ready for merging unless you have plans for more changes.

AdelKS:
It's all good for me, let's spin this and get tons of bug reports! :frog: :laughing:

Tk-Glitch:
> It's all good for me, let's spin this and get tons of bug reports! frog laughing

LeZ GoW :frog:

PR #251, opened 2021-05-31
prepare: Prevent config prompt w/ modprobed-db.

CodingKoopa:
Seeing as #239 nixed the `yes "" |` solution to Tk-Glitch/PKGBUILDS#404, I looked for another way to prevent the prompts. As @Tk-Glitch mentioned that adding the entries wasn't successful, I was convinced that there is a bug in `make localmodconfig`, and stepped through the Perl script for a few hours.

In the end, though, it seems that adding those two config lines fixes the issue just fine. I still don't really understand why this occurs, but this indeed should be better than piping newline key presses to `make`.


Tk-Glitch:
Some work was done on modprobed-db's side as well afaik. That might have fixed the loop. Thank you! :frog: :heart:

PR #256, opened 2021-06-04
Futex2 v4 patchset

lambroisie:
This adds the futex2 v4 patchset to linux 5.13. Please let me know if there is rebasing that needs to be done, or whether this should also be backported to 5.10, 5.11, and 5.12. Thanks.


Tk-Glitch:
I guess you found out that futex2 v3 and v4 patchsets are not compatible with current proton/wine implementations. Thank you for your intent nonetheless!

PR #258, opened 2021-06-06
install.sh: Generic + Clang: prompt to replace libunwind with llvm-libunwind

AdelKS:
Hello,

I need this little thing with my (Gentoo) Generic install with Clang :D 

Thank you!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -191,6 +191,14 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
   # Run init script that is also run in PKGBUILD, it will define some env vars that we will use
   _tkg_initscript
 
+  if [[ "${_compiler}" = "llvm" && "${_distro}" = "Generic" ]]; then
+    read -p "Replace \"libunwind\" with \"llvm-libunwind\" ? y/[n]:" _libunwind_replace
+    if [[ "${_libunwind_replace}" =~ ^(y|yes|Yes|Y)$ ]]; then
+      export LDFLAGS_MODULE="-unwindlib=libunwind"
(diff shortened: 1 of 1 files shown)


CodingKoopa:
A couple of thoughts:
- I take it that there's no `customization.cfg` variable that goes with this, to silence this prompt. Perhaps there should be a variable added for it, or the prompt removed if there's no reason for anyone to *not* want this.
- It might to explain why one might want this option.

Thanks!

Tk-Glitch:
Works for me. Thanks Adel! :frog: :heart:

AdelKS:
Happy to help ! @Tk-Glitch 

Sorry for the late answer @CodingKoopa, so let me answer your two bullet points:
- I thought it would just clutter `customization.cfg` for such an edge case: needing to be in `Generic` on top of using `Clang`, and I made a quick search to see if Fedora/Ubuntu/Archlinux provide `llvm-libunwind` and I couldn't find anything. So yes in Gentoo you need to "manually" swap it out: uninstall regular libunwind then install the llvm drop-in replacement (although it doesn't work with wine for whatever reason). I can query the build system, but only for Gentoo, but the install path I use is `Generic` and not `Gentoo`, maybe soon enough I should add a `Gentoo` specific thing ? Although Gentoo offers so much variety that it just fits with `Generic` and make `Generic` offer many options. I though of making the script generate something that the Gentoo package manager handles, but there's no running away from writing and EBUILD, I think it is not flexible enough to have just one that can install various versions, unfortunately. Although I need to ask.
- It's more of something that if you don't know what it is, don't use it xD I should've written this in the message maybe :thinking:

PR #261, opened 2021-06-06
prepare: restrict user input to digits in _prompt_from_array

AdelKS:
Hello again!

I discovered that entering any text string to prompts from `_prompt_from_array` in `prepare` actually makes it chose something although it shouldn't and should ask again. This should fix it.

Adel


Tk-Glitch:
Nice catch! Thanks Adel :frog: :heart:

PR #265, opened 2021-06-15
Make vmlinux stripping optional 

lordkitsuna:
as the title says this just adds the option to not strip the vmlinux file in case you need debug tools like fadd2line


Tk-Glitch:
Thanks kitsuna! :heart: :frog:

PR #268, opened 2021-06-16
Update README and Adding Gentoo linux-tkg-sources

HougeLangley:
I have added linux-tkg to gentoo, in gentoo-zh overlay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,14 @@ cd void-packages
 ```
 If you have to restart the build for any reason, run `./xbps-src clean linux-tkg` first.
 
+#### Gentoo Linux
+
+sudo emerge --ask app-eselect/eselect-repository
+sudo eselect repository enable gentoo-zh
+You could use `equery u linux-tkg-sources` to see which tweaks would you like to use, the default one is BMQ, if you want to choose others, please using USE to modify.
+sudo emerge --ask sys-kernel/linux-tkg-sources
+The default .config you could find in `/var/db/repos/gentoo-zh/sys-kernel/linux-tkg-sources/config/`
+
 #### Other linux distributions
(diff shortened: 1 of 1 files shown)


HougeLangley:
Sorry for md format not good

AdelKS:
Hello,

I have had a look at your changes here https://github.com/microcai/gentoo-zh/pull/1044/files and the EBUILD is actually a rewrite of what `install.sh` does. Unfortunately, it is not efficient as you need to always modify the EBUILD to keep it up to date with the changes here. Also, there are some modifications in the kernel `.config` file that are done by the [prepare](https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-config/prepare) that is not mirrored in your EBUILD rewrite. Given my current knowledge, `linux-tkg` as-is supports best `Gentoo` with the `Generic` distro approach, as one doesn't need to mirror changes in a parallel EBUILD.

HougeLangley:
OK, thanks for your reply. mirror changes is hard to do.

AdelKS:
Yea, unfortunately. But if you can find a way to leverage the `prepare` script in an EBUILD without the need to keep track with the changes here, that would be awesome! I have been thinking about it for a while and couldn't come up with a proper idea: I think `EBUILDS` just don't fit with linux-tkg... The archlinux `PKGBUILD` uses for example the `prepare`script. But I think `EBUILD`s do not have the right to be interactive. :thinking:

HougeLangley:
Yes, That's right, users could not change anything on `prepare` section. Only use `USE` control which tweaks would them like.

moaalseiari:
> Yes, That's right, users could not change anything on `prepare` section. Only use `USE` control which tweaks would them like.

I saw your ebuild and it looks like it is messing some patches for ondemand aggresive cpu scheduler for intel. 

Btw, I love your youtube channel, you are the real deal! 
I think this project needs more support. I think we can try to find a way to optimize gentoo with more optimized patches. Good luck guys.

moaalseiari:
Also. I wonder why you dont add caclue and stock cpu scheduler. Pds/bmq are not performing well at all.

PR #287, opened 2021-07-27
Changes to Anbox instructions

dr460nf1r3:
I created an `anbox-support` package which takes care of every step needed to get Anbox running without having to do manual changes to the filesystem. -> https://aur.archlinux.org/packages/anbox-support
Also, `anbox-git` recently got some changes - one of them being `/usr/lib/systemd/system/dev-binderfs.mount` being shipped which mounts binderfs without further intervention. It might be good to reflect that in the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ systemctl enable anbox-container-manager.service
 ```
 
 You're set to run Anbox.
-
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thanks! :frog:

PR #302, opened 2021-09-04
Add 5.14 to allowed versions

karx1:
This PR adds 5.14 to the comment above the `_version` option in customization.cfg


Tk-Glitch:
Thanks! :frog:

Issue #296, opened 2021-08-20
deb-pkg failure [Ubuntu, Linux 5.13.12]


Mantas-2155X:
Hello, today I wanted to upgrade my kernel but for some reason deb-pkg fails.

The error and few lines before it happening:
``` 
  LD [M]  drivers/iio/industrialio.o
  CC [M]  drivers/iio/pressure/st_pressure_spi.o
  LD [M]  drivers/iio/pressure/bmp280.o
  LD [M]  drivers/iio/pressure/st_pressure.o
make[2]: *** [debian/rules:7: build-arch] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:77: deb-pkg] Error 2
make: *** [Makefile:1580: deb-pkg] Error 2
-> exit cleanup done
```

<details><summary>customization.cfg</summary>
<p>

```
# linux-TkG config file

# Linux distribution you are using, options are "Arch", "Void", "Ubuntu", "Debian", "Fedora", "Suse", "Generic".
# It is automatically set to "Arch" when using PKGBUILD.
# If left empty, the script will prompt
_distro="Ubuntu"

# Kernel Version - Options are "5.4", "5.7", "5.8", "5.9", "5.10", "5.11", "5.12", "5.13"
_version="5.13"

#### MISC OPTIONS ####

# External config file to use - If the given file exists in path, it will override default config (customization.cfg) - Default is ~/.config/frogminer/linux-tkg.cfg
_EXT_CONFIG_PATH=~/.config/frogminer/linux-tkg.cfg

# [Arch specific] Set to anything else than "true" to limit cleanup operations and keep source and files generated during compilation.
# Default is "true".
_NUKR="true"

# [install.sh specific] Git mirror to use to get the kernel sources, possible values are "googlesource.com" and "kernel.org"
_git_mirror="kernel.org"

# Custom compiler root dirs - Leave empty to use system compilers
# Example: CUSTOM_GCC_PATH="/home/frog/PKGBUILDS/mostlyportable-gcc/gcc-mostlyportable-9.2.0"
CUSTOM_GCC_PATH=""

# Custom LLVM compiler root dirs - Leave empty to use system llvm compiler
# Example: CUSTOM_LLVM_PATH="/home/frog/PKGBUILDS/mostlyportable-llvm/llvm-mostlyportable-11.0.0"
CUSTOM_LLVM_PATH=""

[...]

PR #304, opened 2021-09-05
Improve Generic code path + Gentoo + tmpfs building

AdelKS:
Hello,

So everything is in the title, I did do some small fixes too. I tested on Ubuntu, Fedora and Gentoo. It all seems good!

Adel

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,7 +29,7 @@ plain() {
 _distro_prompt() {
   echo "Which linux distribution are you running ?"
   echo "if it's not on the list, chose the closest one to it: Fedora/Suse for RPM, Ubuntu/Debian for DEB"
-  _prompt_from_array "Debian" "Fedora" "Suse" "Ubuntu" "Generic"
+  _prompt_from_array "Debian" "Fedora" "Suse" "Ubuntu" "Gentoo" "Generic"
   _distro="${_selected_value}"
 }
 
@@ -39,7 +39,7 @@ _install_dependencies() {
   fi
   if [ "$_distro" = "Ubuntu" ]; then
     msg2 "Installing dependencies"
-    sudo apt install git build-essential fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev ${clang_deps} -y
+    sudo apt install git build-essential fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev libelf-dev ${clang_deps} -y
   elif [ "$_distro" = "Debian" ]; then
     msg2 "Installing dependencies"
     sudo apt install git wget build-essential fakeroot libncurses5-dev libssl-dev ccache bison flex qtbase5-dev bc rsync kmod cpio libelf-dev ${clang_deps} -y
@@ -91,14 +91,14 @@ _linux_git_branch_checkout() {
     cd linux-src-git
 
     # Remove "origin" remote if present
-    if git remote -v | grep "origin" ; then
+    if git remote -v | grep -w "origin" ; then
       git remote rm origin
     fi
 
-    if ! git remote -v | grep "kernel.org" ; then
+    if ! git remote -v | grep -w "kernel.org" ; then
       git remote add kernel.org https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
     fi
-    if ! git remote -v | grep "googlesource.com" ; then
+    if ! git remote -v | grep -w "googlesource.com" ; then
       git remote add googlesource.com https://kernel.googlesource.com/pub/scm/linux/kernel/git/stable/linux-stable
     fi
 
@@ -106,13 +106,7 @@ _linux_git_branch_checkout() {
     msg2 "Reseting files to their original state"
 
     git reset --hard HEAD
-
-    if [ "${_distro}" = "Generic" ]; then
-      msg2 "Please enter your sudo password so the script can clean root owned files from the kernel sources"
-      sudo git clean -fdx
-    else
-      git clean -f -d -x
-    fi
+    git clean -f -d -x
   fi
 
   if [[ "$_sub" = rc* ]]; then
@@ -141,7 +135,7 @@ _linux_git_branch_checkout() {
     fi
 
     msg2 "Switching to linux-${_basekernel}.y"
-    if ! git branch --list | grep "linux-${_basekernel}-${_git_mirror}" ; then
+    if ! git branch --list | grep -w "linux-${_basekernel}-${_git_mirror}" ; then
       msg2 "${_basekernel}.y branch doesn't locally exist, shallow cloning..."
       git remote set-branches --add kernel.org linux-${_basekernel}.y
       git remote set-branches --add googlesource.com linux-${_basekernel}.y
@@ -157,7 +151,6 @@ _linux_git_branch_checkout() {
     msg2 "Checking out latest release: ${_kernel_tag}"
     git checkout "${_kernel_tag}"
   fi
-
 }
 
 if [ "$1" != "install" ] && [ "$1" != "config" ] && [ "$1" != "uninstall-help" ]; then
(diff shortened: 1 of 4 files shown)


LecrisUT:
I have confirmed the permission error. At a second installation, right after requesting sudo permission and performing the cleaning of the old linux-src-git, the root usage is maintained when pulling and preparing the next session.

AdelKS:
Thank you for your feedback. Actually, with the new approach of this PR, there should be no root owned files in `linux-src-git` any longer. I will fix that and ask you to confirm :D

LecrisUT:
Quick question, doesn't the makefile support `BUILDDIR` and such to make the cleanup and copying easier?

Also what's the purpose of removing the `*.ko` files after `modules_install` and before `make install`?
https://github.com/Frogging-Family/linux-tkg/blob/a5685a4086b377d4fe70e09b9cf9ed2510ab7340/install.sh#L497-L505
Are these properly ported over to `/lib/modules` during `modules_install` and no longer needed?

AdelKS:
> Quick question, doesn't the makefile support BUILDDIR and such to make the cleanup and copying easier?

I am not aware of this option, if you have the documentation at hand, please refer me to it. Otherwise I will look it up. Maybe it will enable me to write shorter and cleaner code.

> Are these properly ported over to /lib/modules during modules_install and no longer needed?

Yes I think so, my kernel properly boots after that. Therefore `modules_install` actually copies over the modules so they are no longer needed in the original sources, I believe.

I am not the most knowledgeable in these things and feedback is welcome. thanks!

LecrisUT:
Here's what I found about the output directory
https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/Documentation/admin-guide/README.rst#build-directory-for-the-kernel
I guess the following will work. I'll test it on my own as well
```shell
make O=./build
```
Using `make config` or equivalents we should find the other configurable options as well.

LecrisUT:
This part is confusing
https://github.com/torvalds/linux/blob/27151f177827d478508e756c7657273261aaf8a9/Makefile#L1331-L1336
From how I read it, `make install` should automatically call  dkms and install the available modules (probably need to unset `INSTALL_PATH`). I'll double check how the nvidia driver and stuff are installed and if I can manually add them via dkms to the new kernels.

LecrisUT:
A very specific issue for `akmods` needed for nvidia driver
```
akmodsbuild: error: line 31: Invalid version (double separator '-'): 5.14.1-tkg-cfs: Provides:         kernel-modules-for-kernel = 5.14.1-tkg-cfs
```
Seems the naming convention has to be considered

install.sh
```diff
@@ -370,80 +380,155 @@ if [ "$1" = "install" ]; then
...
+    else
+      _kernelname=$_basekernel.${_kernel_subver}-$_kernel_flavor
+    fi
 
-      if [[ "$_sub" = rc* ]]; then
-        _kernelname=$_basekernel.${_kernel_subver}-${_sub}-$_kernel_flavor
-      else
-        _kernelname=$_basekernel.${_kernel_subver}-$_kernel_flavor
+    msg2 "Building kernel"
+    _timed_build make ${llvm_opt} -j ${_thread_num}
```
LecrisUT, line 435: Can use `make O=./build` here, but you also have to add it here
https://github.com/AdelKS/linux-tkg/blob/a5685a4086b377d4fe70e09b9cf9ed2510ab7340/linux-tkg-config/prepare#L1303
and also fix the location of where the `.config` file would be found.

I have not tested if it's needed in other parts after, hopefully it is enough to find them all.

AdelKS:
> My recommendation based on the other formats is [...]

Yea that Fedora quirk is already done in the `Fedora` install path. This is "Generic" and does not have to follow those limitations as other distros do not have that issue. But given, that the call `make rpm-pkg` produces packages that replaces the currently installed kernel, I am thinking of making the Fedora install script use `Generic` and only mod it with this naming quirk for example.

install.sh
```diff
@@ -91,26 +91,26 @@ _linux_git_branch_checkout() {
...
 
     msg2 "Current branch: $(git branch | grep "\*")"
     msg2 "Reseting files to their original state"
 
-    git reset --hard HEAD
-
-    if [ "${_distro}" = "Generic" ]; then
+    if [[ "${_distro}" =~ ^(Generic|Gentoo)$ ]]; then
       msg2 "Please enter your sudo password so the script can clean root owned files from the kernel sources"
+      sudo git reset --hard HEAD
```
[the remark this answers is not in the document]
AdelKS: Yes I will do that. All sudo calls on the folder `linux-src-git` will be removed as the kernel will not be installed from there any longer.

LecrisUT:
> But given, that the call `make rpm-pkg` produces packages that replaces the currently installed kernel

Maybe there is a better solution down the line to not have it delete the old package.

> This is "Generic" and does not have to follow those limitations as other distros do not have that issue

Hmm, would it create issues to other distros though? Does Ubuntu and the likes not follow the same convention? I've caught this issue from the `akmods` usage, so maybe that is the standard that all should use?

install.sh
LecrisUT: Honestly, it would be so awesome if the kernel had a `cmake` compatibility, where all of those standards are baked in. Also `ninja` builds for faster compilation.

install.sh
```diff
@@ -370,80 +380,155 @@ if [ "$1" = "install" ]; then
...
+
+    msg2 "The installation process will run the following commands:"
+    echo "    # copy the patched and compiled sources to /usr/src/$_headers_folder_name"
+    echo "    sudo make modules_install"
+    echo "    sudo make install"
+    echo "    sudo dracut --force --hostonly ${_dracut_options} --kver $_kernelname"
+    echo "    sudo grub-mkconfig -o /boot/grub/grub.cfg"
+
+    msg2 "Note: Uninstalling requires manual intervention, use './install.sh uninstall-help' for more information."
+    read -p "Continue ? Y/[n]: " _continue
```
LecrisUT, line 482: I don't know why, but sometimes the installer just refuses to continue past this point even as I enter `Y`. e.g.:
```
 -> Note: Uninstalling requires manual intervention, use './install.sh uninstall-help' for more information.
Continue ? Y/[n]: Y
 -> exit cleanup done
```
Maybe there's a better failsafe version of the prompt that can trim spaces, invisible characters, etc.

AdelKS:
> Maybe there is a better solution down the line to not have it delete the old package.

Yes, to have the package name changed, `make rpm-pkg` down the line calls `makerpm` or whatever the tool that makes PRMs, and maybe there's a way to give an option to make it overwrite the package name in the spec file. I looked it up for a while then gave up. If you have knowledge in RPM packaging and would like to help there, it would be awesome!

> Hmm, would it create issues to other distros though? Does Ubuntu and the likes not follow the same convention?

The convention is the same except for dashes being replaced by underscores quirk. Dashes are actually broadly used in deb based distros (examples of kernel versions are in #296 like `linux-image-5.11.0-31-generic`) and Arch.

> I've caught this issue from the akmods usage, so maybe that is the standard that all should use?

Fedora afaik is the only distro using akmod ? Arch and Ubuntu use dkms, gentoo it's the package manager that handles module rebuilding.

> Probably the parsing error is because it thinks it's looking at a revision number, so we should at least deliminate with a period at that point.

We just need to not use dashes in Fedora and Suse x)

install.sh
AdelKS: Can you print the value of `_continue` in that `if` block, with quotes (to see white spaces), to see what's wrong ?

install.sh
AdelKS: Well, the kernel is super old and developed its own [build system](https://www.kernel.org/doc/html/latest/kbuild/index.html). Seeing that change is very unlikely.

LecrisUT:
Update on trying to install the nvidia driver, it seems that the `akmods` are even more demanding:
```log
2021/09/09 01:00:40 akmods: Building RPM using the command '/sbin/akmodsbuild --kernels 5.14.1.tkg-pds /usr/src/akmods/nvidia-kmod.latest'
warning: user mockbuild does not exist - using root
warning: group mock does not exist - using root
warning: user mockbuild does not exist - using root
warning: group mock does not exist - using root
error: Package has no %description: kmod-nvidia-5.14.1.tkg-pds
Installing /usr/src/akmods/nvidia-kmod.latest
Building target platforms: x86_64
Building for target x86_64
2021/09/09 01:00:40 akmodsbuild: * Rebuilding /usr/src/akmods/nvidia-kmod.latest for kernel(s) 5.14.1.tkg-pds: warning: user mockbuild does not exist - using root
2021/09/09 01:00:40 akmodsbuild: warning: group mock does not exist - using root
2021/09/09 01:00:40 akmodsbuild: warning: user mockbuild does not exist - using root
2021/09/09 01:00:40 akmodsbuild: warning: group mock does not exist - using root
2021/09/09 01:00:40 akmodsbuild: error: Package has no %description: kmod-nvidia-5.14.1.tkg-pds
2021/09/09 01:00:40 akmodsbuild: Installing /usr/src/akmods/nvidia-kmod.latest
2021/09/09 01:00:40 akmodsbuild: Building target platforms: x86_64
2021/09/09 01:00:40 akmodsbuild: Building for target x86_64
2021/09/09 01:00:40 akmodsbuild: 
2021/09/09 01:00:40 akmods: Building rpms failed; see /var/cache/akmods/nvidia/470.63.01-1-for-5.14.1.tkg-pds.failed.log for details
```

AdelKS:
> Update on trying to install the nvidia driver, it seems that the akmods are even more demanding

It works with the normal approach when using `Fedora` in the script ?

LecrisUT:
I think what the error is referring to is that the kernel is not installed via `rpm` so it does not have metadata such as `%description` to work on.

AdelKS:
I see, so the `Generic` approach for Fedora is out of question. The only hope left is to be able to change the package name. I looked up how to change it _after_ the RPMs are created, but I didn't find any good documentation. Because the actual files that are installed do not conflict with any other kernel. It's just in the spec that it says "this package is 'kernel' and upgrade/downgrades kernel"

install.sh
```diff
@@ -370,80 +372,165 @@ if [ "$1" = "install" ]; then
...
+    msg2 "Removing modules from source folder in /usr/src/${_kernel_src_gentoo}"
+    sudo find . -type f -name '*.ko' -delete
+    sudo find . -type f -name '*.ko.cmd' -delete
+
+    msg2 "Installing kernel"
+    sudo make install
+    msg2 "Creating initramfs"
+    sudo dracut --force --hostonly ${_dracut_options} --kver $_kernelname
+    msg2 "Updating GRUB"
+    sudo grub-mkconfig -o /boot/grub/grub.cfg
```
LecrisUT, line 514: Some distros use `grub2-mkconfig` and similar naming conventions. You also might want to install to `/boot/efi/...` instead. Not sure why the script managed to run on Fedora though.

AdelKS: Well, in any case, the `Generic` approach should not be used by rpm users. Hopefully the other approach of renaming the package names works so this is not needed.

AdelKS:
Renaming the kernel package name in Fedora is not useful as kernels can still be installed in parallel, up to a certain number of installed kernels. So everything is good. I dropped those changes and moved them to [this branch](https://github.com/AdelKS/linux-tkg/tree/fedora-kernel-rename) if ever it's needed in the future.

AdelKS:
Alright, I think we are all good to go @Tk-Glitch

One merge please -frog v2

Tk-Glitch:
Thanks guys! :frog: :heart:

PR #305, opened 2021-09-08
Use Arch config by default + state when config_hardened is used

### LINKED ISSUE #296: deb-pkg failure [Ubuntu, Linux 5.13.12]

AdelKS:
Hello,

This PR makes using Arch's config file default for all distros. I think it is better in many aspects:
- Same experience accross distros
- Avoids those Ubuntu build errors due to signing, this fixes #296 for example
- Avoid the problems non-Arch uses encounter when compiling new kernels on old running kernels where they need to chose what the new options should be.
- Makes it easier to deal with issues, since we know the starting config file is the one we are used to
- If people get issues, they can report it and I can quirk that out or revert this change

Two changes related to this
- Explicitly state that we are using `config_hardened` when it is set
- Add `_configfile="running-kernel"` option to use the config file of the currently running kernel.

This PR is to be merged after #304 since the other PR will contain changes to make the kernel not overwrite the current kernel of RPM distros.

That's it! Feedback welcome!


Tk-Glitch:
I'll need a small rebase and we're good to go

AdelKS:
Here you go! I am a bit tight in time that's why I didn't rebase right away :P

PR #311, opened 2021-09-17
Unify Ubuntu and Debian dependency installation

v1993:
My KUbuntu installation turned out to lack `libelf-dev`, causing build error. I originally wanted to simply add it to Ubuntu package list, but then noticed that Debian option already includes it.

Since there's no harm in trying to install already installed packages, I've just merged two codepaths to use Debian's more extensive package list.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,10 +37,7 @@ _install_dependencies() {
   if [ "$_compiler_name" = "llvm" ]; then
     clang_deps="llvm clang lld"
   fi
-  if [ "$_distro" = "Ubuntu" ]; then
-    msg2 "Installing dependencies"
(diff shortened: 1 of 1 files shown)


AdelKS:
If the dependencies are available in both distros with the same name, then yes it makes us win few lines of code, have you tested both distros ? Which distro did you try otherwise ? For example Fedora and Suse do not name _all_ their packages the same way.

v1993:
I've tested it in Ubuntu, since list remains unchanged for Debian.

AdelKS:
Awesome. Probably the Debian/Ubuntu differentiation is not needed any longer, I will have a look.

Tk-Glitch:
Thanks! :frog: :heart:

PR #313, opened 2021-09-20
Enable support for building a vanilla-esque kernel devoid of patches.

telans:
Hello @Tk-Glitch!

I've made some changes here which when setting `_vanilla_build="true"` in `customization.cfg` disables all default patching mechanisms aside from the cpu opts patch. I did this so I can test different optimizations and issues arising when using LLVM. It's a little easier to narrow things down if there aren't multiple patches potentially introducing issues.

Let me know if you think this should be refactored somewhat or if you have any concerns. Cheers

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -339,7 +339,6 @@ _tkg_patcher() {
 }
 
 _tkg_srcprep() {
-
   if [ "${_distro}" = "Void" ] && [ -e ${srcdir}/sum_failed ]; then
     exit 1
   fi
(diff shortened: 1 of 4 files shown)


linux-tkg-config/prepare
```diff
@@ -1029,7 +1034,7 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\r
 
   # Default to 500Hz if _timer_freq is not set
   if [ -z "$_timer_freq" ]; then
-    _default_index="${_default_timer_frequencies_index[$_cpusched]}"
+    _default_index="${_default_timer_frequencies_index[${_cpusched:-cfs}]}"
```
telans, line 1037: This default needed to be set as default when unset due to `_vanilla_build` skipping `_set_cpu_scheduler()`

AdelKS:
I think it is better to move the entire patching part of the script into a new function. Including the commands that change the config file, into a new function, that will be skipped entirely when `_vanilla_build="true"`. For the scheduler, just overwrite it with `_cpusched="cfs"`

telans:
Yeah, I think you're right.

AdelKS:
If you want to implement it this way, you do not have to make an exception for Graysky's cpu opt patch, it's not worth it. And passing `-march=native` to `KCONFIG` and chosing "core 2" or "amd athlon" or something in make xconfig is strictly equivalent to that patch.

karx1:
Why can't you just compile the kernel from source without patching it? It would achieve the same result

telans:
> Why can't you just compile the kernel from source without patching it? It would achieve the same result

Because I'd like to compile with patches & LLVM to find any issues. Why would I make this easier for myself? :frog:

Besides, a vanilla build option is useful nonetheless for people wishing to use linux-tkg

AdelKS:
@karx1 I thought about it too, and actually linux-tkg helps with many annoying repetitive stuff: ccache, setting up a config file with different defaults, modprobed-db, clang+lto, hadling kernel source code (pulling, cleaning...etc), compiling it, using tmpfs to build. Stuff like that, so it is worth it to enable vanilla-building for those who want it.

PR #314, opened 2021-09-20
prepare: allow forcing LLVM_IAS

telans:
Untested but I'm sure it works :frog:

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -98,6 +98,10 @@ _cpusched=""
 # For advanced users.
 _compiler=""
 
+# Force the use of the LLVM Integrated Assembler whether using LLVM, LTO or not.
+# Set to "1" to enable.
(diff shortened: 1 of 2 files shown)


AdelKS:
Hello, I would actually just go with always having `LLVM_IAS=1` when using `clang`, but I mean why not have a toggle, it offers more flexibility :D

telans:
> Hello, I would actually just go with always having `LLVM_IAS=1` when using `clang`

I'm interesting in making these changes to find issues with patches and builds re ClangBuiltLinux, and sometimes LLVM's IAS causes a couple warnings. It's always good making things toggleable :-)

<sup><sup>linux-tkg ebuild when 🐸</sup></sup>

Tk-Glitch:
Thanks @telans ! :frog:

PR #315, opened 2021-09-20
prepare: allow choosing oldconfig as the .config updating tool

telans:
For manual intervention :frog:


Tk-Glitch:
Thanks telans! :frog:

Issue #327, opened 2021-10-01
Failing to boot on Ubuntu 20.04


Salekin-1169:
I've been using this repo to build kernel for my system (thanks for the great work, really appreciate it)
But it's been failing for me since `5.14` was added (don't remember the exact date, but at the time of fail [this pr](https://github.com/Frogging-Family/linux-tkg/pull/304) wasn't merged yet)

I keep getting `Target filesystem doesn't have requested /sbin/init` error on boot
![err](https://user-images.githubusercontent.com/49670756/135666623-0ce56a40-70a7-4073-9702-6e39456c2314.jpg)
But my pc runs flawlessly when I revert back to the default kernel (from boot menu)

Here's the attached config file. The only changes I made is 1000hz, no hz and runqueue mc
[customization.zip](https://github.com/Frogging-Family/linux-tkg/files/7269157/customization.zip)

I've tried these so far:

`gcc-9, gcc-10, gcc-11` = `Target filesystem doesn't have requested /sbin/init` 

`kernel 5.13, 5.14` = `Target filesystem doesn't have requested /sbin/init` 

`_configfile="running-kernel"` = compilation fails with dpkg error
`_configfile="linux-tkg-config/5.14/config.x86_64"` = compilation finishes, but same error on boot 

Please help me figure out what am I doing wrong here

Spec:
Ubuntu 20.04.3
default kernel (5.11)
everything is on default out of the box

AdelKS:
Can you try compiling with `config-works` from [this issue reply](https://github.com/Frogging-Family/linux-tkg/issues/322#issuecomment-928614736) and see if the resulting kernel boots ? You need to unzip at the same level as `install.sh` and set `_configfile` to that name.

Issue #322, opened 2021-09-26
Launching 5.14.8 goes straight to intramfs screen with no USB keyboard


jhu54:
Had the same problem with 5.14.7 - compile using either bmq or pds, select stock options. rocketlake optimisations and 0 for config. Compile occurs as per normal, but when I select installation, I note that usual error about missing amd firmware does not appear and attempting to load the kernel (5.14.7 or 8) drop into the intramfs command prompt with no USB keyboard. This did not occur with 5.14.6.

Ubuntu target and full tickless were other options, GCC compiler...

netbospl:
To be honest my PC was unstable under kernel 5.14.
It was hanging ofter (every few hours, sometimes less).
I've decided to fallback to 5.13, no hangs for past 72 on 5.13.
I'm happy with 5.13 myself.

chipshopness:
Same here on Ubuntu with both 5.14.7 and 5.14.8.

AdelKS:
Please re-compile from an official ubuntu kernel and use `_configfile="current-kernel"` in `customization.cfg`

jhu54:
Assume you mean _configfile="running-kernel" ...

AdelKS:
Hello @jhu54 ,

I think it is an `initramfs` compression algorithm issue where the kernel is not set to decompress the algorithm used by Ubuntu by default. Can you attach your `/etc/initramfs-tools/initramfs.conf

Can you also please attach the config file of the "5.14.8" kernel that works here ? You can obtain it by running the kernel then doing
```
cp /boot/config-`uname -r` config-works
```

> Strange it did not work with the stock Ubuntu kernel from focal...

It is normal as the kernel is "old" and its config file leads to issues when "upgrading" it to 5.14

ptr1337:
@AdelKS 

I got also one report that my kernel does not work. 

`Hi i found bug in your latest kernel 5.14.8-1-cacule-rdb-llvm

Usb not working propely usb mouse after connect to usb doesn't work and booting show repeating warning about xhci_hcd`

He wrote me. But stock does work. Maybe a patch ? Cant debug it atm.

AdelKS:
Thanks @jhu54 

Unfortunately my initramfs hypothesis is wrong: it is compressed with lz4 and both the default config (that doesn't work for you) and the config that works enable decompressing lz4 initramfs. Is there any error message when the computer drops in initramfs ? That would be helpful. What's your filesystem type, BTRFS ?

[...]

PR #330, opened 2021-10-04
Fix boot issues in Ubuntu (and potentially Debian)

AdelKS:
Thanks to @cpw the boot issues reported in Ubuntu have been figured out!

The issue was that Ubuntu's initramfs generation tool doesn't play nice with zstd compressed modules... Who knew. The fix also involves Debian, although I don't know if it's the same tool or not.

Adel

Closes: #327
Closes: #322

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1302,6 +1302,14 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\r
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thanks @cpw, thanks @AdelKS ! 🐸 ❤️

jhu54:
Awesome thanks for the hard work!!!!

Issue #323, opened 2021-09-26
LLVM compilation on Gentoo fails with Clang error


Slaviusz:
LLVM compilation on Gentoo Linux throws Clang error right away:

I tried also with LLVM/Clang 11.x.x with the same result.

```
#
# merged configuration written to .config (needs make)
#
 -> Setting config
clang-12: error: --rtlib=libgcc requires --unwindlib=libgcc
clang-12: error: --rtlib=libgcc requires --unwindlib=libgcc
make[1]: *** [scripts/Makefile.host:95: scripts/basic/fixdep] Error 1
make: *** [Makefile:553: scripts_basic] Error 2
 -> exit cleanup done
```

owenthewizard:
I have this issue on a new Void Linux install as well.

Slaviusz:
Ok I resolved it. It seems by default USE flags on Gentoo do not include `default-compiler-rt` which is required. I also suggest `default-lld`.
USE flag `llvm-libunwind` is experimental and not supported systemwide as net-fs/samba still depends on `sys-libs/libunwind` which blocks installation of `sys-libs/llvm-libunwind`.

I have created new file inside `/etc/portage/package.use/` and put the line inside:
```
sys-devel/clang -llvm-libunwind default-compiler-rt default-lld
```

AdelKS:
> Ok I resolved it. It seems by default USE flags on Gentoo do not include 1default-compiler-rt` which is required. I also suggest `default-lld`.

Thanks for your feedback! I think even without `default-lld`, `lld` will be used as the compilation is made with `make LLVM=1`, see [here](https://www.kernel.org/doc/html/v5.15-rc2/kbuild/llvm.html#llvm-utilities)

> USE flag llvm-libunwind is experimental and not supported systemwide as net-fs/samba still depends on sys-libs/libunwind which blocks installation of sys-libs/llvm-libunwind.

Yes! It is not needed but I am impatiently waiting for this [bug report`(https://bugs.gentoo.org/791349) to get processed, I am editing by hand `net-fs/samba` to take either `sys-libs/llvm-libunwind` or `sys-libs/libunwind`. I do not use samba, but the change at least makes the thing compile. I would love to just have a use flag for Samba so I can remove it once and for all.

AdelKS:
[...]

PR #332, opened 2021-10-05
README: Gentoo: remove inaccurate dependency on a Clang use flag

AdelKS:
I tested the USE flag change from #323 and in the end it's not needed as I could successfully compile without the `default-compiler-rt` USE flag.


Tk-Glitch:
Thank you Adel! :frog: :heart:

PR #333, opened 2021-10-06
prepare: disable GCC_PLUGINS by default

AdelKS:
Hello,

From what I am seeing about the use of `GCC_PLUGINS` [here](https://www.kernel.org/doc/html/v5.15-rc2/kbuild/gcc-plugins.html) it is not even worth having it enabled. 

What do you think ?

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -667,12 +667,8 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\r
   _module "I2C_NCT6775"
 
   # ccache fix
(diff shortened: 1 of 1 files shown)


ReperakDev:
I think this removes CCache? Don't a lot of people (Gentoo users) use it very frequently? I might be totally wrong though haha

AdelKS:
It's actually the opposite: it helps `ccache` work, the condition I removed is this one `if [ "$_noccache" != "true" ]; ` "noccache" not true x) and it checks if `ccache` is available before disabling `GCC_PLUGINS`

PR #342, opened 2021-11-01
linux513/514/515-tkg: Add futex_waitv() patchset from 5.16 as a toggle option

Iglu47:
Not enabled by default since this make currents public Wine/Proton builds with old fsync interfaces broken.
https://github.com/ValveSoftware/wine/pull/128
https://github.com/andrealmeid/futex_waitv_patches


Tk-Glitch:
Thanks @Iglu47 ! :frog: :heart:

PR #345, opened 2021-11-03
Improve futex_waitv compatibility with Fsync1

Iglu47:
Adds https://gitlab.collabora.com/tonyk/linux/-/commit/b70e738f08403950aa3053c36b98c6b0eeb0eb90 to facilitate the transition.

I didn’t do any performance measurements, I just made sure it works for older Wine builds.
Actually has the same purpose as https://github.com/Frogging-Family/linux-tkg/commit/caad3e526dfdd96dc28e39ee071e9576fffe3a5c

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -501,6 +501,7 @@ case $_basever in
         0007-v5.13-fsync.patch
         0007-v5.13-futex2_interface.patch
         0007-v5.13-futex_waitv.patch
+        0007-v5.13-fsync1_via_futex_waitv.patch
         0007-v5.13-winesync.patch
         0008-5.13-bcachefs.patch
         0009-glitched-ondemand-bmq.patch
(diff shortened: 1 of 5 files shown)


Tk-Glitch:
Thanks, I was going to do it tomorrow :frog:

Iglu47:
some explanations:
`0007-v5.14-futex_waitv.patch` - incompatible with `0007-v5.14-fsync.patch` and `0007-v5.14-futex2_interface.patch`
`0007-v5.14-fsync-waitvcompat.patch` - old checked Fsync1 code using slightly changed function names for compatibility
`0007-v5.14-fsync1_via_futex_waitv.patch` - gives access to the new code through the old Fsync1 interface.

Just apply `0007-v5.14-futex_waitv.patch` with `0007-v5.14-fsync1_via_futex_waitv.patch` or `0007-v5.14-fsync-waitvcompat.patch` to maintain compatibility with Fsync1.
No way to save `0007-v5.14-futex2_interface.patch` at this moment - although it is still possible to rewrite these patches so that they do not conflict, it will require changes on the Wine side, so, as the result, it will not work with older Wine/Proton builds. (I'm sure we'll see more work on this in the future after the kernel 5.16 release).

andrealmeid:
@Iglu47 I would forget about the "futex2_interface" patches, they are tricky to port and should be deprecated in favor of the merged interface.

PR #351, opened 2021-11-08
prepare: Fix and Cleanup some unnecessary checks related to Futex2

Iglu47:
Sorry, it looks like this check does not work properly,
but all these checks can be omitted due to 3462554540a75f346bd65a73ecd458fbfe21da29 in which the behavior of the script has changed.
This way it will simplify the script and fix the problem at the same time.


Tk-Glitch:
Yeah, looks fine like this. Thanks :)

PR #363, opened 2021-11-21
5.15.4

netbospl:
Bumped rel number, updated to newer patch sha256 sum. updated version in linux-tkg-config/prepare

I'm currently building it on Garuda as I send this request, so far, so good.


netbospl:
updated also remaining 5.4.161 5.10.81 5.14.21

Tk-Glitch:
Thanks! :frog:

Issue #366, opened 2021-11-24
Bcachefs is on 5.15 now!


Sukid:
Sorry for yet another pester about this, but bcachefs is on 5.15.* now, could ya add that? :point_right: :point_left: 

(really looking forward to it being mainlined. it's made my own storage system so much more convenient..)

Sukid:
I'm afraid I don't. I definitely will take the time to learn at some point though.

gardotd426:
I've got you covered. I got the patch and am building now (the patch applied cleanly, just want to make sure there are no build errors). I'll push the PR when the build is done.

Sukid:
Thank you!!

PR #367, opened 2021-11-26
Add bcachefs patch for 5.15

gardotd426:
Went ahead and brought up the 5.15 bcachefs patch support as someone was requesting in #366. 

I made sure that the patch both applied cleanly and built properly, it's all good to go.


Tk-Glitch:
Thanks gardotd426! 🐸

gardotd426:
:frog:

Issue #356, opened 2021-11-12
Missing dependency flex on Fedora when running 'config' and using configuration fragments


itspngu:
The logic in `install.sh` explicitly does not call `_install_dependencies` when the user runs the 'config' command: https://github.com/Frogging-Family/linux-tkg/blob/master/install.sh#L198

On my system that borks, as shown in the output below. The 'config' command does not complete successfully when using a configuration fragment as I'm missing missing `flex` (not entirely sure when or how I uninstalled it, upgraded from F34 to F35 the other day). Installing `flex` lets the script do its job - I'm not sure if this is a corner case that can be ignored or if this warrants adjustments to `install.sh`, possibly also for other distros.

<details>
<summary>$ ./install.sh config</summary>

```shell_session
./install.sh config
 -> External configuration file /home/pngu/.config/frogminer/linux-tkg.cfg will be used and will override customization.cfg values.
kernel.org	https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git (fetch)
kernel.org	https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git (push)
googlesource.com	https://kernel.googlesource.com/pub/scm/linux/kernel/git/stable/linux-stable (fetch)
googlesource.com	https://kernel.googlesource.com/pub/scm/linux/kernel/git/stable/linux-stable (push)
 -> Current branch: * (HEAD detached at v5.14.17)
 -> Reseting files to their original state
HEAD is now at 3dfa869cb Linux 5.14.17
Removing .config
Removing .config.old
Removing Documentation/admin-guide/kernel-parameters.txt.orig
Removing Documentation/admin-guide/pm/amd-pstate.rst
Removing Documentation/scheduler/sched-BMQ.txt
Removing Documentation/vm/multigen_lru.rst
Removing arch/arm64/kernel/cpufeature.c.orig
Removing arch/x86/Makefile.orig
Removing block/elevator.c.orig
Removing drivers/Makefile.orig
Removing drivers/cpufreq/Kconfig.orig
Removing drivers/cpufreq/amd-pstate.c
Removing drivers/cpufreq/cpufreq_ondemand.c.orig
Removing drivers/cpufreq/intel_pstate.c.orig
Removing drivers/firmware/efi/secureboot.c
Removing drivers/gpu/drm/amd/amdgpu/amdgpu_drv.c.orig
Removing drivers/gpu/drm/amd/amdgpu/amdgpu_ttm.c.orig
[...]

PR #370, opened 2021-11-27
README: Generic Install: state that deps need to be manually installed

AdelKS:
Hello,

Following #356, a little update to the README to clarify that dependencies need to be manually installed for the generic install.

Adel


Tk-Glitch:
Thanks Adel! :frog:

PR #373, opened 2021-11-29
5.16.rc3

pongo1231:
Updated version + sha256 sum and bumped rel number.

Built just fine on my system.


ptr1337:
Are you sure that the kernel built correctly ? yesterday i built the tkg-rc kernel and every module failed at building several modules (not only nvidia one)

i already found for my pkgbuilds how to solve it.

pongo1231:
The kernel by itself is working fine, however you are right that modules won't build on any of the 5.16 releases as of now. As a workaround I simply overrided the `Makefile.modfinal` file with the same file fetched from a 5.15 kernel (in my case from `/usr/lib/modules/5.15.4-221-tkg-pds-llvm/build/scripts`) which lets me build them just fine.

ptr1337:
Yes, need to check the PKGBUILD of it since a _package-headers change is needed but only for 5.16. 

Watching about it later.

Tk-Glitch:
Thanks! :frog:

PR #389, opened 2021-12-19
Add support for 5.15.10 and 5.16.rc6

gardotd426:
Compiles and installs cleanly.


gardotd426:
Added 5.16.rc6 as well. Running it right now, with the new v6 AMD_CPPC patch, which I'm about to file a PR for over at community-patches

Tk-Glitch:
Thanks gardotd426! :frog: :heart:

PR #396, opened 2022-01-12
Update PKGBUILD

devacom:
Update to 5.4.171 5.10.91 5.15.14 5.16.0


Snoop05:
just FIY you also need to change versions in linux-tkg-config/prepare

ninelore:
Maybe you should merge your `patch-3` and `patch-4` branches and create a PR with both the changed PKGBUILD and prepare

PR #399, opened 2022-01-12
Added schedtool to dependencies installation

fiordiconio:


Tk-Glitch:
We don't want to enforce schedtool installation. We used to do it and there were complaints from people who didn't want to install/use it, so we made it optional.

fiordiconio:
Okay, thank you for the clarification

PR #391, opened 2021-12-30
Kernel versions update

netbospl:

PR #401, opened 2022-01-17
update kernels

devacom:
kernels 5.16.1 5.15.5 5.10.92 5.4.172


Tk-Glitch:
I have merged https://github.com/Frogging-Family/linux-tkg/pull/391 which was serving the same purpose. Thank you very much !

PR #408, opened 2022-01-21
add ebuild

qwe795138426:


Tk-Glitch:
wth

itspngu:
0 commits :monocle_face:

Snoop05:
delet dis, install windows xp instead :joy:

AdelKS:
Ebuilds cannot bundle the flexibility of linux-tkg. These are missing the entirety of the `.config` tweaks. Also, one can install only one kernel 5.x version at a time. Although it may be suitable for some Gentoo users.

qwe795138426:
oh, bot forced pulled and covered my change

qwe795138426:
> Ebuilds cannot bundle the flexibility of linux-tkg. These are missing the entirety of the `.config` tweaks. Also, one can install only one kernel 5.x version at a time. Although it may be suitable for some Gentoo users.

Of course, I don't think a full `.config` is necessary because I think most of Gentoo users are more likely to decide their config by themselves. 

Through USE, it might be not influence its flexibility. To people in need, applying patches by themselves rather than use a pre-configured sources maybe more suitable.

If you like, I can add `.config` as defconfig to deal with it.

qwe795138426:
Let me try once more.

PR #412, opened 2022-01-27
add ebuild

qwe795138426:


qwe795138426:
It might works

pog5:
Gentoo support already exists.
```
cd /usr
sudo git clone https://github.com/Frogging-Family/linux-tkg
cd linux-tkg
bash install.sh
```

Issue #410, opened 2022-01-26
Cacule Repository Archived, Active Fork


Orangian:
Hi,
I noticed that the CacULE scheduler was not available for 5.16. It appears that the original repository was archived, however, someone else has maintained it [here](https://github.com/CachyOS/cacule-cpu-scheduler). If possible, would you all consider using this new fork of CacULE to support it on newer Linux versions?

ptr1337:
It was available since 5.15. Also the orginal creator of cacule is back in the team and maintain the cacule patches there.

Orangian:
Ah, I see. I only did a cursory glance of that repository (and the situation as a whole) so I didn't realize that. If that is the case, is it possible for  linux-tkg integrate CacULE for the more recent versions, or is there another issue?

ptr1337:
No there is no issue. Maybe add it to the community patches or just use it at userpatch :)

Orangian:
Alrighty, thanks!

Orangian:
Digging deeper into this is very interesting. In commit  [#2d603642dab724bf2c059560ab0cf0bc0aa5b1bc](https://github.com/Frogging-Family/linux-tkg/commit/2d603642dab724bf2c059560ab0cf0bc0aa5b1bc) it appears that TKG attempted to allow for its use above 5.15, however, when actually going to do so, the script does not properly run, and says that CacULE is not available for anything above 5.15. I think this is due to lines 164-167 in  [linux-tkg/linux-tkg-config/prepare](https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-config/prepare), which contain the following: 

`  elif [ "$_basever" = "516" ]; then
    _avail_cpu_scheds=("pds" "bmq" "cfs")
  elif [ "$_basever" = "517" ]; then
    _avail_cpu_scheds=("pds" "bmq" "cfs")`
Marking cacule as being unavailable for 5.16 or 5.17. In addition, line 21 in the README.md should be updated from 
`- CacULE by Hamad Marri: [code repository](https://github.com/hamadmarri/cacule-cpu-scheduler)`
to
`- CacULE by Hamad Marri: [code repository](https://github.com/CachyOS/cacule-cpu-scheduler)`

I'll make a pull request to fix these issues.

PR #414, opened 2022-01-27
Fix CacULE 5.15+ and CacULE link in README

Orangian:
Hi, 
In this pull request I updated linux-tkg/linux-tkg-config/prepare to make CacULE available on 5.16 and 5.17. A few lines in prepare will need to be updated as CacULE releases patches for newer kernels, there may be a better way to do this.

I also updated line 21 in the README to properly link people to the new CacULE repository. In Issue #410, ptr1337 stated that the original creator of CacULE is working on patches there so there does not appear to be a need to also link the original repository.  (I can remove this change from the pull request if requested.)


Orangian:
Just saw that TK Glitch already fixed this, closing.

PR #417, opened 2022-02-02
Bump mainline kernel to 5.16.5.

Edu4rdSHL:
Signed-off-by: Edu4rdSHL <edu4rdshl@protonmail.com>

PR #418, opened 2022-02-03
Kernel Version Bump

netbospl:
Kernel Version Bump, updated sha256sum's, pkgrel bump


Tk-Glitch:
Thanks netbospl! Also fixes https://github.com/Frogging-Family/linux-tkg/pull/417

PR #429, opened 2022-02-10
5.16.8 5.15.22 5.10.99 5.4.178

devacom:


Tk-Glitch:
Thank you! As this is a duplicate of https://github.com/Frogging-Family/linux-tkg/pull/426 (with the addition of pkgrel bump), I'll merge the older PR. Since defconfig update is needed, I'll bump the pkgrel with that instead.

PR #432, opened 2022-02-11
Implement Github workflow to automatically bump kernel versions

AdelKS:
Hello!

I finally finished this work. So here's how this works:
- There's a script at the root of the repo, called `update-kernel-versions.sh`, that anyone can use on his own local copy of `linux-tkg` to bump versions automatically
  - It uses `git ls-remote` to fetch latest tags, and sources the `prepare` script to compare whats on the repo and what's upstream
  - If upstream offers a newer version, it changes
    - The `prepare` script with the newer numbers
    - Updates `PKGBUILD` sha sums by downloading both the older and newer archives to know which sha sum to replace with the newer one
      - rc-to-release changes are also taken care of, you need to always put the patch shasum on an rc kernel to `#upcoming_kernel_patch_sha256` (I was out of ideas for a better naming) so it can be found and replaced with the correct shasum
      - 5.xx (initial release) to 5.xx.y versions changes are also taken care of
  - It writes a file called `kernel_updates` where the kernel version bumps are written
- The  `update-kernel-versions.sh` script is then run by a workflow, every six hours (can also be triggered by hand). that will use the `kernel_updates` to push a commit with the correct information, where I am the author (so I take the glory for ALL kernel version bumps :smiling_imp: ) and FROGMINER, if that's okay (making only  FROGMINER as an author makes the commit message preview ugly on Github because the name appears twice).

I tested it and it works ! :dagger: 

Adel

### DIFF
--- a/.github/workflows/auto-update-kernel-versions.yml
+++ b/.github/workflows/auto-update-kernel-versions.yml
@@ -1,0 +1,21 @@
+name: Automatic version bumping
+on:
+  schedule:
+    - cron: '3 */6 * * *'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          git config user.name "BIG UGLY FROGMINER"
+          git config user.email BIG_UGLY_FROGMINER@froggi.es
+          ./update-kernel-versions.sh
+          changes=$(cat kernel_updates)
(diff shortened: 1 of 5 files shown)


Tk-Glitch:
Thanks Adel! Let's do it. :frog: :frog: :frog:

AdelKS:
Ohoooo I'm hyped! Click on that "Run workflow" button now !!!

Issue #352, opened 2021-11-09
Ubuntu/Debian Install Script Error (unary operator expected)


Hexcede:
On line 348 of the `install.sh` script (head is at commit db405096bd7fb52656fc53f7c5ee87e7fe2f99c9 for future reference) there is an error during the install prompt on my system. Even though its just a single line, I don't full well understand what the issue is because I only recently started using linux & bash and haven't familiarized myself with the syntax here yet.

Contents of lines 347-349:
```bash
read -p "Do you want to install the new Kernel ? Y/[n]: " _install
if [[ $_install =~ [yY] ]] || [ $_install = "yes" ] || [ $_install = "Yes" ]; then
  cd "$_where"
```

I'm not confident but I'm fairly sure it has to do with the `yY` selection.

AdelKS:
Hello!

Change it to this:
```shell
if [[ $_install =~ ^(y|Y|yes|Yes)$ ]]; then
```

Does it fix the issue you are facing ?

Hexcede:
> Hello!
> 
> If you change it to this:
> 
> ```shell
> if [[ $_install =~ ^(y|Y|yes|Yes)$ ]]; then
> ```
> 
> Does it fix the issue you are facing ?

It seems so, yes. I am unsure if the command executed as there wasn't any output, but, no script errors.
![image](https://user-images.githubusercontent.com/8394472/141035110-d247e8e8-88d3-473b-b18f-303603a51a26.png)

AdelKS:
Run `./install.sh uninstall-help` to see if the new kernel you installed is listed.

Hexcede:
> Run `./install.sh uninstall-help` to see if the new kernel you installed is listed.

It looks that way, but, I had manually installed the kernel before, based on the script. The build I just produced is the same as the one I already had installed.

I'll do another build later for something I don't have installed and see.

AdelKS:
You mean you installed it from the `DEBS` folder. That's good enough.

Hexcede:
> Hello!
> 
> If you change it to this:
> 
> ```shell
> if [[ $_install =~ ^(y|Y|yes|Yes)$ ]]; then
> ```
> 
> Does it fix the issue you are facing ?

Hey I apologize for the late response, I just wanted to mention that this does indeed fix the issue, however this does have one problem.

The option will select yes by default normally, but the statement you provided doesn't do that (it explicitly checks for yes, whereas it should actually explicitly check for no I believe)

[...]

PR #434, opened 2022-02-12
install.sh: quote every variable used in if statements

AdelKS:
Small fixes in `install.sh`

Closes: #352


Tk-Glitch:
Thanks! :frog: :heart:

PR #439, opened 2022-02-14
Fix cpu-opt patch in "install.sh" and make "prepare" verbose about it

### LINKED ISSUE #438: Additional CPU families patch does not apply anymore?
I've noticed after building 5.16.8 kernel with BMQ  scheduler that CPU family is set to `generic-x86` - probably as a falback after `zen2` option was not found when applying config from my `custom.myfrag` file. I had a look and it seems the additional CPU family patches are not being applied, at least to 5.16.latest. I've been searching commit messages but did not find a mention of this patch removal.

Is this a bug/regression?

Thanks for your help.

AdelKS:
This is the fix #438, with a small update to `install.sh` to have fewer lines of code and I also made `prepare` verbose about graysky's patches so we notice earlier if something is off.

Adel.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -210,36 +210,16 @@ if [ "$1" = "install" ] || [ "$1" = "config" ]; then
 
   cd "$_where"
 
-  case "$_basever" in
-    "57")
+  if [ "$_basever" = "54" ]; then
+    opt_ver="4.19-5.4"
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Thanks Adel! 🐸 ❤️

Issue #440, opened 2022-02-15
linux-tkg not installing on arch linux


Abhs15:
Hello i have just installed arch linux and wheni try and install this tkg-pds it gives me and error like this 
An error was encountered applying patches. It was logged to the prepare.log file.
  -> Applying Glitched PDS patch
/home/genos/Downloads/git/linux-tkg/linux-tkg-config/prepare: line 385: patch: command not found

i have installed tkg-pds before on my previous setup and it worked fine its was very fast too so what seems to be the problem here is there a dependency i have to install to fix this i have missed?
here is the log file

AdelKS:
you need to install the `patch` package

Snoop05:
Whole `base-devel` group in fact :)

AdelKS:
> Whole base-devel group in fact :)

that cannot be added to dependencies in the PKGBUILD right ?

Abhs15:
tnx so much guys worked

netbospl:
> > Whole base-devel group in fact :)
> 
> that cannot be added to dependencies in the PKGBUILD right ?

Groups of packages like base-devel wouldn't work by adding it, all packages in the group would have to be added one by one. But all packages in this group could potentially be added but to be honest, base-devel group packages are in some sense 'must have to be able to build packages' at least from my linux daily driving experience so far.

PR #441, opened 2022-02-15
PKGBUILD: add 'patch' as a dependency

AdelKS:
Closes: #440


Snoop05:
I dont think that issue is a valid bug tbh, if you are building packages using `makepkg` you are supposed to have `base-devel` group installed which includes `patch`

in general, runtime dependencies that are in `base` group don't go to depends array, and same for `base-devel` and makedepends

AdelKS:
Awesome, so he probably manually deleted the package or something.

PR #444, opened 2022-02-19
uninstall-help for Debian

zecktos:
use the Ubuntu uninstall-help also when "_distro" is set to Debian like in the installation part


Tk-Glitch:
Yup, should be allright. Thanks zecktos! :frog: :heart:

PR #463, opened 2022-03-15
Fixes for the recent cpu uarch changes

openglfreak:
Fixes the following error:

    /usr/share/makepkg/util/message.sh: line 69: printf: `B': invalid format character

and the patch not being applied because of a wrong file name.


AdelKS:
Oh haha my first version of wget was with -O then I thought I'd make it shorter x) thanks!

Tk-Glitch:
Oh yeah, that old error :frog:
Thanks openglfreak ! :heart:

PR #473, opened 2022-03-24
Fixed broken checksum for `patch-5.4.187.xz`

alou-S:
The sha256sum for the patch was wrong hence preventing the PKGBUILD to build when choosing version 5.4


Tk-Glitch:
Thank you alou-S! :frog: :heart:

PR #476, opened 2022-03-24
linux-5.17: bump kernel .config from upstream Arch

AdelKS:
Hopefully this will fix the issues with 5.17 kernel people are reporting

File here: https://github.com/archlinux/svntogit-packages/blob/packages/linux/trunk/config


Tk-Glitch:
Yup, that was coming. Thanks for doing it before I did :frog: :heart:

PR #477, opened 2022-03-24
Fix version bump script: what a silly typo...

AdelKS:
After I figured out the issue... I reacted like this "oh god... that's the reason ??!!"

Yea...


Tk-Glitch:
We've all been there :D Thanks Adel! :frog: :heart:

PR #478, opened 2022-03-25
Bump pkgrel.

Edu4rdSHL:
It will allow automated builds to rebuild the kernels with the recent commits, which will probably fix some issues, especially since the kernel config was changed.


Tk-Glitch:
Sure. Thanks! 🐸

PR #480, opened 2022-03-25
Updated README.md to reflect DKMS Clang / llvm support

flindeberg:
Updated the README to better suit that DKMS v3.0.2 and later handle clang / llvn out of the box.

Since DKMS v3.0.2 still is quite recent perhaps the warning should stay a bit longer.


Tk-Glitch:
Thank you flindeberg! :frog: :heart:

PR #490, opened 2022-04-06
linux-5.16: add bcachefs patch

AdelKS:
A little PR to please @Sukid xD

I haven't tried it yet, please confirm that all is good.

Adel

PR #494, opened 2022-04-09
fix checksum for 5.16 after 2bd3a4af9132b3329906c6d8201fb3ebc89356a7

Iglu47:
small checksum fix for https://github.com/Frogging-Family/linux-tkg/pull/490

sorry, this shift happened because I didn't know what the commented out hashsum refers to when the fsync/winesync patches were added.


Tk-Glitch:
Weird I didn't notice.. I even have a built package. Thanks! 🐸

Issue #493, opened 2022-04-09
5.16 and 5.17 failed to build


wamuu-sudo:
![unknown](https://user-images.githubusercontent.com/92715832/162561957-2fac7c2b-438f-428f-beca-0a4e80effd71.png)

Iglu47:
hmm, strange. it looks like https://gitlab.com/alfredchen/linux-prjc/-/issues/55
and it really should affect the 5.17 and 5.16 branches (no 5.15 and lower) 
but 5.17 at least should be fixed by https://github.com/Frogging-Family/linux-tkg/commit/30be4c4bc586f2253a040112c2241db42c8d577e#diff-2bc2b8a9bf96f086b7aab42dc938438f8276804df5484fe4ca9d1531d9c64022R9934

wamuu-sudo:
5.16 and 5.17 still not working for me

Iglu47:
well, at least there is clearly the issue with 5.16 in upstream (which apparently has not been reported yet), but I think that this simple fix from 5.17 @Tk-Glitch will quickly add even before waiting for an official fix.

Iglu47:
@wamuu-sudo what build environment/distribution are you using?

wamuu-sudo:
Gentoo of course, with running-kernel parameter.

wamuu-sudo:
5.17.2-cfs works properly

wamuu-sudo:
I've recompile and 5.17.2-pds works now.
I'll check 5.16.19

Iglu47:
> I'll check 5.16.19

probably 5.16 will be built only after the fix from the PR above.

PR #495, opened 2022-04-09
linux516: Fix Project C compilation

Iglu47:
Fix build fail due to multiple definition of `node_reclaim_distance`.
part of: https://github.com/Frogging-Family/linux-tkg/commit/30be4c4bc586f2253a040112c2241db42c8d577e

Fixes: https://github.com/Frogging-Family/linux-tkg/issues/493


Iglu47:
it would probably be nice to check not needed for 5.16 second part https://github.com/Frogging-Family/linux-tkg/commit/30be4c4bc586f2253a040112c2241db42c8d577e

Iglu47:
uff, seems `init/Kconfig` part of https://github.com/Frogging-Family/linux-tkg/commit/30be4c4bc586f2253a040112c2241db42c8d577e requires not only for 5.16, but also probably for lower versions

Tk-Glitch:
Thanks, yeah looks like we'll want to port the changes over. :frog:

PR #500, opened 2022-04-16
Bug reports: create "logs" folder and populate it at each run

AdelKS:
This makes it easier for people to open issues as we will ask them to attach every file in the "logs" folder that contains the logs for only the last build : 
-  a copy of `customization.cfg` that were used for the last build
- `prepare.log` 
-  the full output in the shell
   - in `install.sh`, this is achieved with the `script` command.
   - I couldn't firgure a seamless way for Arch people: to obtain the shell-output.log, we will ask them to do it by hand: `makepkg [args] | tee shell-output.log`
- system information (distro, version of `clang` and `gcc`), am I missing something here ?
these files have the `.txt` extension so they can easily be drag & dropped into Github issues.

Please make an issue template (I do not have permissions to do that) that:
- Asks people to describe their issue and ALWAYS attach the contents of their `logs` folder by drag & dropping each file into their issue message. And a special note for Arch users to always run `makepkg [args] | tee shell-output.log` (or to do it only when they get issues).
- Asks people to check for relevant keywords in the issues search bar, including closed issues, and maybe the solution is there already.

A little welcome QoL improvement :dagger:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ src/*
 */pkg/
 pkg/*
 */linux-5*/
+*.myfrag
+logs/*
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+# If current run is not using 'script' for logging, do it
+if [ -z "$SCRIPT" ]; then
+  export SCRIPT=1
+  /usr/bin/script -q -e -c "$0 $@" shell-output.log
+  exit_status="$?"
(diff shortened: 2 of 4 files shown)


Tk-Glitch:
Template was added with https://github.com/Frogging-Family/linux-tkg/commit/749d68a7e97f78498f744cde1bba6805f3ed4ebb

Thanks Adel! :frog: :heart:
Let's merge!

Issue #508, opened 2022-05-03
Commit 08b1f7 (Update winesync patches) breaks LLVM compilation


Slaviusz:
**OS:** Gentoo Linux amd_64
**Description:** With current master/head (commit 08b1f7) compilation with LLVM 14.0.3 (probably other LLVM versions as well) compilation fails with an error (see below)
Compilation with GCC works and kernel boots.
```
drivers/misc/winesync.c:667:9: error: variable 'ret' is uninitialized when used here [-Werror,-Wuninitialized]
        return ret;
               ^~~
drivers/misc/winesync.c:648:9: note: initialize the variable 'ret' to silence this warning
        int ret;
               ^
                = 0
1 error generated.
make[2]: *** [scripts/Makefile.build:288: drivers/misc/winesync.o] Error 1
make[1]: *** [scripts/Makefile.build:550: drivers/misc] Error 2
```
Related to #503

Iglu47:
@Slaviusz thanks for report ❤

PR #509, opened 2022-05-03
Fix winesync patches

Iglu47:
small update to fix uninitialized variable

Fixes https://github.com/Frogging-Family/linux-tkg/issues/508


Tk-Glitch:
Thanks Iglu47! :frog: :heart:

PR #514, opened 2022-05-15
corrected sha256sum for patch-5.17.8.xz

redgloboli:
seems there was the wrong sha256sum

https://cdn.kernel.org/pub/linux/kernel/v5.x/sha256sums.asc


Tk-Glitch:
Thanks redgloboli. 5.15 was also frogged.

PR #515, opened 2022-05-16
update-kernel-versions.sh: Utilize sha256sums.asc

Iglu47:
Partially replace downloading original files to checking with kernel.org checksum autosigner and sha256sums.asc

addition to https://github.com/Frogging-Family/linux-tkg/pull/432
adds a GnuPG dependency, but as far as I understand the current Github workflow should support this:
https://github.com/actions/virtual-environments/blob/ubuntu20/20220508.1/images/linux/Ubuntu2004-Readme.md

### DIFF
--- a/update-kernel-versions.sh
+++ b/update-kernel-versions.sh
@@ -15,9 +15,53 @@ kernel_tags=$(git -c 'versionsort.suffix=-' \
 source linux-tkg-config/prepare
 trap - EXIT
 
+# Start by making sure our GnuPG environment is sane
+if [[ ! -x /usr/bin/gpg ]]; then
+  echo "Could not find gpg"
+  exit 4
+fi
+
+## Generate the keyring
+if [ ! -s gnupg/keyring.gpg ]; then
+  if [[ ! -d gnupg ]]; then
+    echo "gnupg directory does not exist"
+    mkdir -p -m 0700 gnupg
+  fi
+  echo "Making sure we have all the necessary keys"
+  gpg --batch --quiet --homedir gnupg --auto-key-locate wkd --locate-keys torvalds@kernel.org gregkh@kernel.org autosigner@kernel.org
+  if [[ $? != "0" ]]; then
+    echo "FAILED to retrieve keys"
+    exit 3
+  fi
+  gpg --batch --homedir gnupg --export torvalds@kernel.org gregkh@kernel.org autosigner@kernel.org > gnupg/keyring.gpg
+fi
+
+# Cleanup
+rm -f v*.x.sha256sums{,.asc}
+
 updates=""
 for _key in "${_current_kernels[@]}"; do
   latest_full_ver=$(echo "$kernel_tags" | grep -F "v$_key" | tail -1 | cut -c2-)
+  kver_major="$(echo ${_key} | cut -d. -f1)"
+  kver_base="$(echo ${_key} | tr -d ".")"
+
+  ## Getting sha256sums by sha256sums.asc
+  if [ ! -s v${kver_major}.x.sha256sums ]; then
+    echo "Downloading the checksums file for linux-v${kver_major}.x"
+    curl -sL -o "v${kver_major}.x.sha256sums.asc" https://cdn.kernel.org/pub/linux/kernel/v${kver_major}.x/sha256sums.asc
+    if [[ $? != "0" ]]; then
+      echo "FAILED to download the v${kver_major}.x checksums file"
+      exit 3
+    fi
+    echo "Verifying the v${kver_major}.x checksums file"
+    count_gpg=$(gpg --homedir gnupg --keyring=gnupg/keyring.gpg --status-fd=1 v${kver_major}.x.sha256sums.asc | grep -c -E '^\[GNUPG:\] (GOODSIG|VALIDSIG)')
+    if [[ ${count_gpg} -lt 2 ]]; then
+      echo "FAILED to verify the v${kver_major}.x.sha256sums file."
+      rm -f "v${kver_major}.x.sha256sums"
+      exit 3
+    fi
+    rm -f "v${kver_major}.x.sha256sums.asc"
+  fi
 
   _from_rc_to_release="false"
   if [[ "${_kver_subver_map[$_key]}" == rc* ]]; then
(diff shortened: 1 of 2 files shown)


Iglu47:
- I returned the missing single quotes that are needed for `#upcoming_kernel_patch_sha256`.
- for old RC kernels, hashsums are taken directly from the PKGBUILD now (I suppose the probability of finding the correct values there is very high, and if this is not - they will still be overwritten as an obsolete value)
- for the remaining option, I separated the download from the pipeline

I'm not sure if this script works, because with these changes need to test all options...

> Another thing, I believe you fail when the signature verification fails, maybe it's better to retry downloading both the file and signature like 3 times before failing and exiting the script?

well, could probably expect something like this when working with gpg, but if that's going to be a problem, it's probably better to use some other way to manually manage the generation/delivery of keyring.gpg (especially since they don't update very often).

AdelKS:
> for old RC kernels, hashsums are taken directly from the PKGBUILD now (I suppose the probability of finding the correct values there is very high, and if this is not - they will still be overwritten as an obsolete value)

Very good idea

> for the remaining option, I separated the download from the pipeline

:+1: 

> I'm not sure if this script works, because with these changes need to test all options...

The only option I did not test is going from rc to release. And that one is hard to test (without some extra wokr), we just need to wait for the actual release and we'll see xD

> well, could probably expect something like this when working with gpg, but if that's going to be a problem, it's probably better to use some other way to manually manage the generation/delivery of keyring.gpg (especially since they don't update very often).

I wasn't talking about that, I will comment on the actual line

update-kernel-versions.sh
```diff
@@ -15,9 +15,53 @@ kernel_tags=$(git -c 'versionsort.suffix=-' \
...
+  if [ ! -s v${kver_major}.x.sha256sums ]; then
+    echo "Downloading the checksums file for linux-v${kver_major}.x"
+    curl -sL -o "v${kver_major}.x.sha256sums.asc" https://cdn.kernel.org/pub/linux/kernel/v${kver_major}.x/sha256sums.asc
+    if [[ $? != "0" ]]; then
+      echo "FAILED to download the v${kver_major}.x checksums file"
+      exit 3
+    fi
+    echo "Verifying the v${kver_major}.x checksums file"
+    count_gpg=$(gpg --homedir gnupg --keyring=gnupg/keyring.gpg --status-fd=1 v${kver_major}.x.sha256sums.asc | grep -c -E '^\[GNUPG:\] (GOODSIG|VALIDSIG)')
+    if [[ ${count_gpg} -lt 2 ]]; then
```
AdelKS, line 58: Here, if it fails, I think it's better to re-download the shasum file and its signature again, then test again. 

But you know what ? Let's see if it will actually fail in the upcoming updates. If it fails too frequently, we can do the change. Also, the CI run is run every six hours, so if it fails, in the next run it probably won't. But people will have to wait 6 more hours :fearful: xD

AdelKS:
I tested your updated script on 599eb50 (by checking out that commit and simply copying your updated script as `update-kernel-versions-new.sh`. And I get this issue:

```shell
adel@leyantoo /media/personal/Documents/Gethings/linux-tkg                                                                                                      [21:39:21]
> $ ./update-kernel-versions-new.sh                                                                                                                           [±599eb50 ●]
gnupg directory does not exist
Making sure we have all the necessary keys
pub   rsa4096 2013-01-24 [SC]
      B8868C80BA62A1FFFAF5FDA9632D3A06589DA6B1
uid           [ unknown] Kernel.org checksum autosigner <autosigner@kernel.org>

pub   rsa4096 2011-09-23 [SC]
      647F28654894E3BD457199BE38DBBDC86092693E
uid           [ unknown] Greg Kroah-Hartman <gregkh@kernel.org>
sub   rsa4096 2011-09-23 [E]

pub   rsa2048 2011-09-20 [SC]
      ABAF11C65A2970B130ABE3C479BE3E4300411886
uid           [ unknown] Linus Torvalds <torvalds@kernel.org>
sub   rsa2048 2011-09-20 [E]

Downloading the checksums file for linux-v5.x
Verifying the v5.x checksums file
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
gpg: Signature made Sun 15 May 2022 08:33:07 PM CEST
gpg:                using RSA key 632D3A06589DA6B1
gpg: Good signature from "Kernel.org checksum autosigner <autosigner@kernel.org>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: B886 8C80 BA62 A1FF FAF5  FDA9 632D 3A06 589D A6B1
current version on repository 5.18.rc5 -> 5
upstream version 5.18-rc7 -> 7
Newer upstream
Updating rc kernel version in linux-tkg-config/prepare
Downloading the GZ tarball for linux-5.18-rc5
Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.
FAILED to download the linux-5.18-rc5.tar.gz
```

Iglu47:
> Downloading the GZ tarball for linux-5.18-rc5
> Warning: Binary output can mess up your terminal. Use "--output -" to tell 
> Warning: curl to output it to your terminal anyway, or consider "--output 
> Warning: <FILE>" to save to a file.
> FAILED to download the linux-5.18-rc5.tar.gz

thanks, I didn't notice how it was leaked to the public version. fixed now.

update-kernel-versions.sh
Iglu47: Hmm, Yea, I would come back to this part if there will be any often problem with this. And maybe would make the script re-invoke via Github workflow somehow. So that all exits from the script with error 3 are simply given a second chance by re-running the script.

Iglu47: about 6 hours: Also, I have not yet found a way to replace the scheduled launch of the script. I thought there would be some ready-made option for Github, but maybe something can be done with tracking release tags, for example via email or something, and then triggering an event by writing a comment in the issue/discossion.

AdelKS: Let's merge this PR then we can think of this improvement in another

AdelKS:
Looks all good to me! We can merge this already @Tk-Glitch

AdelKS (approved):

AdelKS:
Unfortunately I still do not have write access to merge PRs so we have to wait for TKG.

Iglu47:
@AdelKS I changed the command for finding `old_kernel_shasum` to be shorter and more correct.

Tk-Glitch:
Looks good to me. Let's give it a spin!
Thank you Iglu47 for diving into it, and thank you Adel for reviewing sand testing the changes!
🐸 ❤️

PR #517, opened 2022-05-18
prepare: move 5.16 to EOL

Iglu47:
move 5.16 to EOL list


Tk-Glitch:
Thanks Iglu47! That was planned for later today 🐸

PR #518, opened 2022-05-21
Predictable alphabetically sorted user fragment application

Slaviusz:
With more than one user kernel config fragment (*.myfrag files) the application of fragments is seemingly random based on find command output making incremental deliberate config fragment application impossible.

This PR adds ascending alphabetical sort operation on top of the find result (array of found files) when searching for user fragments so that fragments are later applied to default kernbel config in alphabetical order.


Tk-Glitch:
Works for me. Thanks Slaviusz! :frog:

PR #519, opened 2022-05-23
Fix hash for 5.18

RodoMa92:
The hash for linux-5.18.tar.xz is incorrect, added the correct hash.


Tk-Glitch:
Thank you RodoMa92! 🐸 ❤️

PR #520, opened 2022-05-23
update-kernel-versions.sh: Improve _from_rc_to_release

Iglu47:
https://github.com/Frogging-Family/linux-tkg/commit/79e921e86acfc28e6c1ca73d555ef289b788f166 did not update hashsum in PKGBUILD (https://github.com/Frogging-Family/linux-tkg/pull/519)
this patch addition to https://github.com/Frogging-Family/linux-tkg/pull/515 and should help in RC to Release upgrade cases, but `sha256sums.asc` has not yet been updated properly.

- back to file download when some hashsums are missing in `sha256sums.asc`
- add `--retry 2` for all curl download commands
- output more information about the hashsums being replaced (allows to immediately see their absence or wrong values)
- remove unnecessary `cat` commands

@AdelKS sorry, looks like I made the script even uglier, but it should work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@
 *.xz
 *.pkg
 *.bak
+*.tar
+*.tar.sign
 *.tar.gz
 *.pkg.*
 *.old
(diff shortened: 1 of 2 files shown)


Iglu47:
- added info message

@AdelKS signature verification is not for tar.xz, should this be left for greater reliability?

Iglu47:
please ignore this PR

PR #522, opened 2022-05-23
update-kernel-versions.sh: Depends more on sha256sums.asc

Iglu47:
Prefer to take a list of kernels tags for stables and longterm branches from `sha256sums.asc` instead git

This patch addition to https://github.com/Frogging-Family/linux-tkg/pull/515 and should solve the hashsums replacement issue when the script execution falls at a time while the kernel.org archive files are not yet available, but the list of tags from git is already fresh
(like https://github.com/Frogging-Family/linux-tkg/commit/79e921e86acfc28e6c1ca73d555ef289b788f166 did not update hashsum in PKGBUILD and https://github.com/Frogging-Family/linux-tkg/pull/519)

- prefer to take the list of kernels tags from `sha256sums.asc` as a guarantee that the files are available for download
- add `--retry 2` for all curl download commands
- output more information about the hashsums being replaced (allows to immediately see their absence or wrong values)
- remove unnecessary `cat` commands

### DIFF
--- a/update-kernel-versions.sh
+++ b/update-kernel-versions.sh
@@ -8,7 +8,7 @@ escape() {
   _escaped=$(printf '%s\n' "$1" | sed -e 's/[]\/$*.^[]/\\&/g')
 }
 
-kernel_tags=$(git -c 'versionsort.suffix=-' \
+kernel_git_tags=$(git -c 'versionsort.suffix=-' \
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Yeah, that looks more robust indeed. Thank you Iglu47! :heart: :frog:

PR #524, opened 2022-05-24
Add support for TaskType Scheduler

### LINKED ISSUE #502: Question: Is it possible to add support for TT
With Cacule's future being uncertain at best, would it be possible to add the successor to Cacule, TaskType? It's from the [same developer](https://github.com/hamadmarri/TT-CPU-Scheduler) behind Cacule and even has the same guy, @ptr1337, maintaining it (since Hamadmarri only maintains it officially for LTS kernels).

I would presume since TT is the successor to Cacule the goal for the scheduler is the same as Cacule and thus would be a good place in this kernel.

Daedalusspacegames:
Enables use of the [TaskType Scheduler](https://github.com/hamadmarri/TT-CPU-Scheduler) with linux-tkg, utilising patches from the CachyOS project (by @ptr1337). 

I have successfully built 5.17 and 5.18 with the TT scheduler, and am currently running TT 5.18 with no issues.

Fixes #502

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -172,6 +172,7 @@ _set_cpu_scheduler() {
     ["muqss"]="MuQSS (Multiple Queue Skiplist Scheduler)"
     ["upds"]="Undead PDS (TkG)"
     ["cacule"]="CacULE"
+    ["tt"]="TT (TaskType)"
   )
 
   # CPU SCHED selector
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
I wanted to give it a try myself before taking a decision about offering support for it, but I couldn't find the time. I guess it won't hurt to try and see. If it ends up making no sense we can always go back :frog:
Thank you @Daedalusspacegames ! :frog: :heart:

RyanHakurei:
@Tk-Glitch I've been patching it in manually since late 5.17, I have not faced any issues and while gaming (Hitman 2, Doom Eternal, 7 Days to Die, just to name a few) I've had no hiccups.

PR #527, opened 2022-05-29
README: notify about issues in DEB and RPM based distros

AdelKS:
Like this users are informed of the current issues and can decide in trying `linux-tkg` with that in mind.


Tk-Glitch:
Thanks Adel! Definitely useful notes. :frog: :heart:

Issue #528, opened 2022-05-30
5.17.xxx TT Build issue


nutterpc:
Found as i go to build a secondary kernel to have on my system i can swap between, and noticed that TT refuses to build:

 -> Switching to linux-5.17.y
  linux-5.17-kernel.org
 -> 5.17.y branch exists locally, updating...
Switched to branch 'linux-5.17-kernel.org'
Your branch is up to date with 'kernel.org/linux-5.17.y'.
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Unpacking objects: 100% (1/1), 808 bytes | 808.00 KiB/s, done.
HEAD is now at e960d7349 Linux 5.17.11
 -> Checking out latest release: v5.17.11
Note: switching to 'v5.17.11'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at e960d7349 Linux 5.17.11
 -> Setting version...
 -> Using Arch patches
--2022-05-30 14:06:55--  https://raw.githubusercontent.com/graysky2/kernel_compiler_patch/master/more-uarches-for-kernel-5.17%2B.patch
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.110.133, 185.199.109.133, 185.199.108.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.110.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 25738 (25K) [text/plain]
Saving to: '/home/nutterpc/linux-tkg/more-uarches-for-kernel-5.17%2B.patch’

/home/nutterpc/linux-tkg/more-uarches-for-kernel-5. 100%[===================================================================================================================>]  25.13K  --.-KB/s    in 0.008s  

2022-05-30 14:06:55 (3.17 MB/s) - '/home/nutterpc/linux-tkg/more-uarches-for-kernel-5.17%2B.patch’ saved [25738/25738]

 -> Applying graysky's cpu opts patch
 -> Applying clear linux patches
[...]

PR #529, opened 2022-05-30
Replace instances of `--output-dir` from curl commands

Daedalusspacegames:
Given that the `--output-dir` was only argument was added in cURL version 7.73.0 (released October 14th, 2020), Ubuntu 20.04 and distributions based on it (among others, I'm sure) do not yet have this argument in the version of cURL that they ship.

This pull replaces the `--output-dir` argument with shell redirection to enable use on the aforementioned distros.

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -615,20 +615,20 @@ _tkg_srcprep() {
     _msg="Applying TT patch"
     if [ "${_distro}" = "Void" ]; then
       if [[ $_basever = 515 ]]; then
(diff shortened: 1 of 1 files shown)


Daedalusspacegames:
Thanks to @nutterpc for exposing this issue (see #528)

nutterpc:
It's not the first time I've helped find a bug in this lovely utility :)

Tk-Glitch:
Thank you @Daedalusspacegames ! :frog: :heart:

Issue #531, opened 2022-05-30
5.18-tkg-cacule hangs on boot


undeadbydawn:
tested on 5.18.0 and 5.18.1. Boot hangs on Loading initial ramdisk

I'm using greysky zen3 (5600X) optimisation, all other options at config default
kernel functions perfectly using cfs

OneOfOne:
I can confirm.

Tk-Glitch:
It's a specific issue with cacule. Since we're not providing the patchset, we'll disable explicit support for it in the 5.18 submenu while waiting for a fix upstream.

ptr1337:
Yes, cacule is currently broken. Checking for a fix.

undeadbydawn:
Cool. I'm currently testing TT, which is working perfectly - so thanks for adding that option

undeadbydawn:
New updated patch still won't launch.
I've mentioned this on the patch thread, posting here for the record

ptr1337:
> New updated patch still won't launch. I've mentioned this on the patch thread, posting here for the record
I did compiled linux-cachyos-cacule cacule-rdb cacule, and every kernel is booting.

Patches which were used can be found here:
https://github.com/CachyOS/linux-cachyos/blob/master/linux-cachyos-cacule/PKGBUILD#L169

If youre on arch, you can add the keyring of the cachyos repo and then just download the kernel, install and test it. I got several from users that its fine now.
Also I do personally currently run cacule:

```❯ uname -a
Linux ptr1337-cachyos 5.18.1-4-cachyos-cacule-lto #1 SMP PREEMPT_DYNAMIC Thu, 02 Jun 2022 05:07:32 +0000 x86_64 GNU/Linux```

regards.

ptr1337:
Its fixed. I did not watched at the commit to the correct kernel patch link, missed to update the cacule-scheduler repo and just had in my kernel-patches repo updates.
https://github.com/CachyOS/cacule-cpu-scheduler/commit/6430b698ef233ee6df205ae509d8a9bc5099c99b

undeadbydawn:
> Its fixed. I did not watched at the commit to the correct kernel patch link, missed to update the cacule-scheduler repo and just had in my kernel-patches repo updates. [CachyOS/cacule-cpu-scheduler@6430b69](https://github.com/CachyOS/cacule-cpu-scheduler/commit/6430b698ef233ee6df205ae509d8a9bc5099c99b)

Confirmed. Kernel now launches.
[...]

PR #534, opened 2022-06-02
Revert "Disable cacule availability on 5.18. The patchset results in …

ptr1337:
…non-booting kernels currently."

This reverts commit 2b225b594ad1bd2b644a2e9c107a4be232942ab7.

fixed cacULE-5.18 patchset with commit https://github.com/ptr1337/kernel-patches/commit/2adb46deccaf3c80dd8e94c87e9904981e20c459

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -199,7 +199,7 @@ _set_cpu_scheduler() {
   elif [ "$_basever" = "517" ]; then
     _avail_cpu_scheds=("pds" "bmq" "cacule" "tt" "cfs")
   elif [ "$_basever" = "518" ]; then
-    _avail_cpu_scheds=("pds" "bmq" "tt" "cfs")
+    _avail_cpu_scheds=("pds" "bmq" "cacule" "tt" "cfs")
   else
     _avail_cpu_scheds=("cfs")
   fi


Tk-Glitch:
Thank you for cacule users ptr1337! :frog: :heart:

undeadbydawn:
Just tried a fresh compile, my cacule kernel still isn't booting.
Back to cfs for now (which is performing unusually well)

AdelKS:
I have been using CFS for nearly a year (?) and I sincerely believe there's no need for other schedulers, at least everyone should try it first, then others if it doesn't work.

ptr1337:
Actually its weird that it is not booting at you, or with the tkg pkgbuild.
Ill do a compile with linux-tkg again. 
Already did mention here some things:
https://github.com/Frogging-Family/linux-tkg/issues/531#issuecomment-1146207248

Tk-Glitch:
While cacule is pretty terrible, I cannot agree for the others that I have tested and used extensively. Namely BMQ and PDS. While the older PDS implementation (undead, which is now fully dead) is still the best gaming CPU scheduler by far (sadly), both BMQ and PDS offer much better latency than CFS.

It's incomparably better now than it was before 5.12, don't get me wrong, but it's still throughput focused. It'll give good benchmark results for sure, but interactivity isn't the best. One could argue that windows 10/11 feels much more reactive than Linux from a latency standpoint, and that's still pretty true for audio and inputs at the very least. 

I'll be happy to claim we don't need anything else than CFS when I'll feel like we're there, but that's not yet true imho. We'll get there though.

ptr1337:
Its fixed. I did not watched at the commit to the correct kernel patch link, missed to update the cacule-scheduler repo and just had in my kernel-patches repo updates

undeadbydawn:
I used pds and bmq for a while, but found they caused weird Steam issues - specifically, games refusing to launch on fresh install. They'd launch on the mainline kernel, then no problem back on tkg-pds/bmq afterwards, but _never_ on first try. I have absolutely no idea why, but it was 100% consistent
I jumped to CacULE purely cos I got some extremely helpful advice from the dev on Reddit, and found it very responsive with no Steam weirdnesss. RDB was godawful, though.
Recent experimenting (non-scientific, purely eyeballing) got me better results with cfs than pds, bmq or TT.
Given that 5.18 brings AMD specific improvements (I'm running 5600X/5700XT), cfs seems a safe bet for now.

undeadbydawn:
> Its fixed. I did not watched at the commit to the correct kernel patch link, missed to update the cacule-scheduler repo and just had in my kernel-patches repo updates

Ah, cool. Glad that's been caught now.
I'll reroll and confirm full function

Issue #532, opened 2022-06-01
[Feature Request] Higher HZ Patch from TT


seed3165:
https://github.com/hamadmarri/TT-CPU-Scheduler/blob/master/patches/5.15/high-hz.patch

PR #538, opened 2022-06-09
Enable TT for 5.19rc and TT High HZ patch for 5.15

Daedalusspacegames:
Allows users to select the TaskType scheduler for the 5.19rc (tested to build and boot successfully).

Also allows enabling the High HZ patch for 5.15 in the config (solves #532). I do not personally use this patch, but I figure there is no harm in giving the option to those who want it (again, able to build and boot this configuration successfully).

If changes should be required for the implementation of the TT High HZ patch, I am happy to make them upon request.

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -201,7 +201,7 @@ _cacule_rdb="false"
 _cacule_rdb_interval="19"
 
 # TT only - Enable High HZ patch (available for 5.15 only) - Default is "false"
-_tt_high_hz="true"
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
The high HZ patch looks pretty horrible to me, and having it only for 5.15 is kinda.. funny, I guess? Someone asked for it, so it's their problem I suppose. Thank you Daedalusspacegames! :frog: :heart:

PR #541, opened 2022-06-16
prepare: move 5.17 to EOL list

Iglu47:
move 5.17 branch to EOL list


Tk-Glitch:
Thanks Iglu47! :frog:

PR #552, opened 2022-07-12
update-kernel-versions.sh: Print last modified timestamp for remote sha256sums.asc

Iglu47:
optional addition to https://github.com/Frogging-Family/linux-tkg/pull/522 for shows more information.
(just to make sure that fresh `sha256sums.asc` was not available during script execution.)

- print last modified timestamp for remote `sha256sums.asc`
- print the current time at the download `sha256sums.asc`
- remove verbose lines


Tk-Glitch:
I'm fine with that. Thanks Iglu47! :frog: :heart:

Iglu47:
(theoretically this should always just duplicate the signature creation time that gpg output shows)

PR #554, opened 2022-07-22
update bcachefs for 5.18, add bcachefs for 5.19, include module-fix

ptr1337:
- Update bcachefs for 5.18
- Add bcachefs for 5.19
- Include module compilation fix, but still recommend to built it with =y right now
 
### DIFF
--- a/linux-tkg-patches/5.19/0008-5.19-bcachefs.patch
+++ b/linux-tkg-patches/5.19/0008-5.19-bcachefs.patch
@@ -1,7 +1,7 @@
-From 98a9e9a069c0619986de099d587dae0158d82eac Mon Sep 17 00:00:00 2001
+From db2079ca2ce5000c3dfe3656c1c7f580b053a325 Mon Sep 17 00:00:00 2001
 From: Peter Jung <admin@ptr1337.dev>
-Date: Fri, 22 Jul 2022 14:22:01 +0200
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
That's for non-MGLRU, right?

ptr1337:
> That's for non-MGLRU, right?

Yes. I have for MG-LRU in my kernel-patches repo. If you want I can add the patches also here and update them every 4 weeks or something like this ?

You can find them here:
5.18: https://github.com/ptr1337/kernel-patches/blob/master/5.18/misc/0001-bcachefs-after-lru.patch
5.19: https://github.com/ptr1337/kernel-patches/blob/master/5.19/0001-bcachefs-after-lru.patch

Tk-Glitch:
There are plans for offering mglru by default soon-ish, so that shouldn't be a problem for long. Thank you ptr1337! 
:frog: :heart:

ptr1337:
> There are plans for offering mglru by default soon-ish, so that shouldn't be a problem for long. Thank you ptr1337! frog heart

Would recommend it anyways. It is really well tested over the latest kernel versions and will be maybe merged with 5.20 anyways (so far there seems nothing to be in linux-next). 

Just a little entry in the customization.cfg to enabled by default or not, so it can be patched all time in and also then enabled manually if the user not used to enable it as default.

PR #556, opened 2022-07-22
update bcachefs sums, enable bcachfs for 5.19

ptr1337:


ptr1337:
Now its fine x)

Tk-Glitch:
I was only going to do it in a few hours from now so thanks ! :frog:

PR #564, opened 2022-08-05
update bcachefs for 5.19, fixes error in compile

ptr1337:
Fixes current issue with the compilation with bcachefs enabled in 5.19

You could also referrer to these patches in the PKGBUILD, these are regularity updated, also the bcachefs-after-lru can be used then if lru is enabled:
https://github.com/ptr1337/kernel-patches/tree/master/5.19/misc
  

Tk-Glitch:
Thanks ptr1337!
I saw the report on Discord and wanted to update the patchset. That's nice of you.

PR #570, opened 2022-08-15
update-kernel-versions.sh: Avoid sha256sums checks at new major RC

Iglu47:
sha256sums.arc is not generated for new major RC

fixes: https://github.com/Frogging-Family/linux-tkg/runs/7843819575?check_suite_focus=true#step:3:19

this gives an extra line from `cut`, but should work fine.
it also looks like `curl` returns 0 if the download fails =\ , so the previous checks don't work properly.


Tk-Glitch:
Yup, that should do. Thanks Iglu47! :frog: :heart:

PR #583, opened 2022-09-01
linux60: Fix Winesync patchset for 6.0-rc3

Iglu47:
small change for smooth applying


Tk-Glitch:
Thanks! :frog: :heart:

compicat:
can somebody confirm winesync is working with newer wine versions than 6.10 which had been hold?

Iglu47:
> can somebody confirm winesync is working with newer wine versions than 6.10 which had been hold?

no, previous versions of fastsync is deprecated and should be avoided, 4 version of winesync and fastsync are currently relevant:
- all current patches in linux-tkg provide 4 version of winesync
- on wine-tkg side, patches are available only for ~7.7+. To support lower versions of Wine, need additionally to backport the actual version of fastsync (which would be a non-trivial task).

compicat:
sorry for thanking late. I should at least say that, winesync is better than futex_wait even it is very little to scale, thanks for this fix and all for your help.

PR #591, opened 2022-09-11
use git for kernel sources & improved CI workflow

AdelKS:
Hello!

Per #525, I finally took the time to work on implementing the transfer: use git to fetch/update kernel sources for all distros, while also forcing to clone `linux-tkg` through `git` so we can make a simpler `PKGBUILD`.     

One thing remains, and that is to properly handle the use of `tmpfs` in Arch. We can either make the entry in `customization.cfg` apply to Arch too, or respect the `BUILDDIR` variable in `/etc/makepkg.conf`. I have no strong opinion on this.

Also, needs testing, too.

Adel


AdelKS:
For the script that updates the dictionary of latest kernel versions, I forgot to update it: we need it now only to trigger CI kernel builds. I will do that.

linux-tkg-config/prepare
```diff
@@ -116,24 +132,69 @@ _prompt_from_array() {
...
+
+  echo -e "_basekernel='$_basekernel'\n_basever='$_basever'\n_sub='$_sub'\n_kver='$_kver'" > "$_where"/BIG_UGLY_FROGMINER
+}
+
 _set_kernel_version() {
 
-  if [ -z "$_version" ] || ! [[  "${!_kver_subver_map[*]}" =~ "$_version" ]]; then
+  # if $_version is a valid version, define $_kernel_git_tag directly
+  # from $_kver_latest_tags_map. Otherwise, empty it so it gets prompted
+  if [[ -n "$_version" && -v _kver_latest_tags_map[$_version] ]]; then
```
AdelKS, line 175: Oh there's an error here, I need to fix it.

linux-tkg-config/prepare
```diff
@@ -179,33 +234,33 @@ _set_cpu_scheduler() {
...
+  if [ "$_kver" = "504" ]; then
     _avail_cpu_scheds=("pds" "bmq" "muqss" "cacule" "cfs")
-  elif [ "$_basever" = "57" ]; then
+  elif [ "$_kver" = "507" ]; then
     _avail_cpu_scheds=("pds" "bmq" "muqss" "cfs")
-  elif [ "$_basever" = "58" ]; then
+  elif [ "$_kver" = "508" ]; then
     _avail_cpu_scheds=("upds" "pds" "bmq" "cfs")
-  elif [[ "$_basever" =~ ^(59|511)$ ]]; then
+  elif [[ "$_kver" =~ ^(59|511)$ ]]; then
```
AdelKS, line 243: need to change `59` to `509`

Ann1kaB:
when i try to change the minor version on kernel 6.0 it doesn't take effect

AdelKS:
Thanks for giving it a try @Ann1kaB 

If you tried changing the numbers in the variable `_kver_subver_map` yes it is not followed anymore. Only `_version` from `customization.cfg` is followed now. I'd need to make it accept kernel git tags to choose subversions. I'll do that I think, even though pkgrel versions bumps sometimes adapt patches to some specific newer kernel subversions

AdelKS:
Okay now the CI should be working ! Here's what it does now:

- periodically checks for new kernel versions and compare it to `.github/latest-built-kernel`
  - if there's a newer one, builds PDS, CFS and BMQ while leveraging `ccache`
    - if the build succeeds, makes a release and pushes the latest built kernel git tag to the file `.github/latest-built-kernel`
- the CI is run only for `master`

 You can check that all of this work well on my fork on master : https://github.com/AdelKS/linux-tkg

AdelKS:
@Ann1kaB now you should be able to set specific kernel versions in `customization.cfg` or as an env variable.

Tk-Glitch:
This is looking pretty good. Regarding tmpfs building on Arch, respecting makepkg config sounds like the most sensible choice to me right now, but that can be discussed later on if needed. Anything you need/want to add before merging?

AdelKS:
Thanks ! 

> but that can be discussed later on if needed. 

Yeah I'd rather deal with that in a follow-up PR if that's okay.

> Anything you need/want to add before merging?

Yeah I need to de-factorize the workflow by removing the YAML anchors and parallelize the build jobs. I should be done with that soon. I'll ping you :D

AdelKS:
@Tk-Glitch 

One merge please -frog v2

Tk-Glitch:
Let's GOOOOOOO !

Thanks @AdelKS for the large update. Also thanks to @Ann1kaB for testing.

:frog: :frog: :frog: :frog: :frog: :frog: :frog:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

AdelKS:
Happy to make `linux-tkg` better for everyone ! :heart: :frog:

PR #593, opened 2022-09-16
Support `doas` and `su` for root authentication

amarakon:
I ran the script and many moments later, I was disappointed because I got the error `sudo: command not found`. This is because I use `doas` instead. Some people use `doas` or `su` instead of `sudo` so they do not have it installed. This change should automatically detect if the user has sudo installed, and if it is not, check if doas is installed, then check if su is installed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,6 +17,14 @@ set -e
 _where=`pwd`
 srcdir="$_where"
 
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Works for me. Thanks @amarakon ! :frog: :heart:

PR #598, opened 2022-09-18
prepare: move 5.18 to EOL list

Iglu47:
move 5.18 branch to EOL list


Tk-Glitch:
Thanks Iglu47. We'll annoy Adel a bit but I see there are other conflicts already so let's go for it. :frog:

PR #605, opened 2022-09-27
Pathing for diffconfig in prepare

jnines:
Relevant to changes in 59afa4083c1dc59b6e26fe0579b41b92575fbf55

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1764,7 +1764,7 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\r
       else (
         prev_pwd="${PWD:-$(pwd)}"
         cd "$_where"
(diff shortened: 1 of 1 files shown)


AdelKS:
Can you please run a quick `./install.sh config` (use generic as distro) and see if doing xconfig works too ? Sorry can't test for now

jnines:
In a Fedora VM it broke, so I'm assuming this is specific to Arch.  Tweaked it to check for Arch/Pkgbuild, was successful on both Arch and Fedora afterwards.  

I'm not a shell scripting guru so if I missed something I apologize.

Tk-Glitch:
It only seems to break through the PKGBUILD from my testing. So yeah, assuming it's specific to it.

Thanks @jnines , nice catch :frog: :heart:

PR #613, opened 2022-10-03
Fix RPM build with Linux 6.0

ktheticdev:
Made a patch that allows Linux 6.0 to be built on Fedora.


Tk-Glitch:
Interesting. Thank you very much !
@AdelKS

AdelKS:
Thanks @ktheticdev, I see that you remade a patch from the same patch I did in my previous PR, but with more context lines. It's better to have less context lines so the patch applies still even with minor changes upstream. 

@Tk-Glitch  it's just better to copy the patch as-is from e.g. the 5.19 patches folder @Tk-Glitch and close this PR :D 

Thanks again @ktheticdev

PR #616, opened 2022-10-04
fix mkspec patches

ktheticdev:
Current mkspec patches do not make `kernel` package provide `kernel-uname-r` package, which is e. g. necessary to build nvidia kmods manually.


Tk-Glitch:
Nice. Weird it wasn't spotted during testing. Thanks! :frog: :heart:

AdelKS:
> Nice. Weird it wasn't spotted during testing.

`akmod-nvidia` compiled without any warnings or errors, thought it was good xD

PR #622, opened 2022-10-09
Add TaskType to Linux 6.0

RyanHakurei:
It builds and boots just fine. 

image

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -297,7 +297,7 @@ _set_cpu_scheduler() {
   elif [ "$_kver" = "519" ]; then
     _avail_cpu_scheds=("cfs" "pds" "bmq" "cacule" "tt" "bore")
   elif [ "$_kver" = "600" ]; then
-    _avail_cpu_scheds=("cfs" "pds" "bmq" "bore")
+    _avail_cpu_scheds=("cfs" "pds" "bmq" "tt" "bore")
   else
     _avail_cpu_scheds=("cfs")
   fi


ptr1337:
Please check the new README of the changes in TT v4. 
He has implemented the cfs balancer into TT, which can be used with the balancer option = 2 (should be default).

Also other latency based sysctl's can be changed and tested

linux-tkg-config/prepare
```diff
@@ -806,19 +806,19 @@ _tkg_srcprep() {
     if [ "${_distro}" = "Void" ]; then
       if [[ $_kver = 515 ]]; then
         curl "https://raw.githubusercontent.com/CachyOS/kernel-patches/master/${_basekernel}/TT/0001-tt-${_basekernel}.patch" > "$srcdir"/tt.patch
-      elif [[ $_kver = 517 ]] || [[ $_kver = 519 ]]; then
-        curl "https://raw.githubusercontent.com/ptr1337/kernel-patches/master/${_basekernel}/sched/0001-tt.patch" > "$wrksrc"/tt.patch
+      elif [[ $_kver = 517 ]] || [[ $_kver = 600 ]]; then
```
Tk-Glitch, line 809: 5.19 also requires `/sched/0001-tt.patch` path. Replacing it with 6.0 here redirects it to the else statement below, breaking 5.19 support.

linux-tkg-config/prepare
```diff
@@ -806,19 +806,19 @@ _tkg_srcprep() {
...
-        curl "https://raw.githubusercontent.com/ptr1337/kernel-patches/master/${_basekernel}/sched/0001-tt-${_basekernel}.patch" > "$wrksrc"/tt.patch
+        curl "https://raw.githubusercontent.com/CachyOS/kernel-patches/master/${_basekernel}/sched/0001-tt-${_basekernel}.patch" > "$wrksrc"/tt.patch
       fi
       tkgpatch="$wrksrc/tt.patch" && _tkg_patcher
     else
       if [[ $_kver = 515 ]]; then
          curl "https://raw.githubusercontent.com/CachyOS/kernel-patches/master/${_basekernel}/TT/0001-tt-${_basekernel}.patch" > "$srcdir"/tt.patch
-      elif [[ $_kver = 517 ]] || [[ $_kver = 519 ]]; then
-         curl "https://raw.githubusercontent.com/ptr1337/kernel-patches/master/${_basekernel}/sched/0001-tt.patch" > "$srcdir"/tt.patch
+      elif [[ $_kver = 517 ]] || [[ $_kver = 600 ]]; then
```
Tk-Glitch, line 818: Same thing as above regarding 5.19.

linux-tkg-config/prepare
Tk-Glitch: This is not I was talking about. Both 5.19 and 6.0 are using an unversioned `0001-tt.patch` (each in their own dir), unlike for example 5.18's patch which is `0001-tt-5.18.patch`. The change above would make the 5.19 kernel try to use a `0001-tt-5.19.patch` which doesn't exist.

RyanHakurei:
I'll work on the issues this evening so that 5.19 support for TT doesn't break with this change.

linux-tkg-config/prepare
ptr1337: > This is not I was talking about. Both 5.19 and 6.0 are using an unversioned `0001-tt.patch` (each in their own dir), unlike for example 5.18's patch which is `0001-tt-5.18.patch`. The change above would make the 5.19 kernel try to use a `0001-tt-5.19.patch` which doesn't exist.

If it is suiting better for you, I can simply rename the patches to "0001-tt.patch". Would be no problem, since we are not using that patches anymore and just keep them updated, if there is any update.

Like moving all current 0001-tt-$basekernel to 0001-tt.patch in each major dir. Just give me a ok, then I will change them

Tk-Glitch: I'm fine with that. It would indeed make things easier, allowing to drop those conditional versioning checks.

ptr1337: > I'm fine with that. It would indeed make things easier, allowing to drop those conditional versioning checks.

https://github.com/CachyOS/kernel-patches/commit/e65beabf2e307d74cd24f5d351d89857830baadd

I hope I did not forget anything. I did also updated the bore scheduler, so that there is just $major/sched/0001-bore.patch is there and can be simply fetched with the $major.
If I missed something, just hit me up.

Tk-Glitch: Thank you very much!

Tk-Glitch:
Following ptr1337 URL changes, I updated your changes. Let's merge :)
Thanks both ! :frog: :heart:

PR #636, opened 2022-10-31
Clarify the description of the tickless parameter

ventureoo:
The linux-tkg has a `_tickless` parameter which controls how the scheduling-clock interrupts behaves in the kernel. There is a Full dynticks mode is described as "Can provide better performance in many cases". The problem is that it is not specified here that without passed `nohz_full` kernel parameter it works the same way as Idle dynticks which is described in kernel option description. Even so, Full dynticks only makes sense when there is one single task running on the processor core (see https://docs.kernel.org/timers/no_hz.html?highlight=tickless#omit-scheduling-clock-ticks-for-cpus-with-only-one-runnable-task) and there are no other tasks to switch. I think this case may be relevant only when isolating CPUs, but not for desktop tasks or games, but I could be wrong.

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -144,7 +144,9 @@ _numadisable="false"
 _misc_adds="true"
 
 # Set to "0" for periodic ticks, "1" to use CattaRappa mode (enabling full tickless) and "2" for tickless idle only.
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Thanks ventureoo ! :frog: :heart:

PR #641, opened 2022-11-06
Update bcachefs patches for 5.15 and 6.0

ptr1337:


Tk-Glitch:
Thanks ptr1337! :frog: :heart:

PR #651, opened 2022-11-19
Update tovic9 prjc sched to latest changes:

ViNi-Arco:
Hey, I'm updating the Torvic9 prjc fork, which contains synchronization to the latest kernel with a possible solution for instability..

### DIFF
--- a/linux-tkg-patches/6.0/0005-glitched-pds.patch
+++ b/linux-tkg-patches/6.0/0005-glitched-pds.patch
@@ -88,3 +88,54 @@ index 9270a4370d54..30d01e647417 100644
  
  static void set_task_reclaim_state(struct task_struct *task,
  				   struct reclaim_state *rs)
+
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Why adding changes to "glitched" patches since it seems to be shared? Wouldn't it make more sense to add it to the prjc patchset?

ViNi-Arco:
> Why adding changes to "glitched" patches since it seems to be shared? Wouldn't it make more sense to add it to the prjc patchset?

Hi Mr.Etienne, I did it this way because Torvic9 did it separately, [here](https://gitlab.com/torvic9/linux60-vd/-/blob/master/prjc-6.0-r0-vd-fix.patch), if this patch causes more problems than it helps it is easier to remove it in "Glitched".

ViNi-Arco:
I include the fix inside the patch #652, hope it helps

PR #658, opened 2022-12-05
Update CPU opt lists to include raptorlake, meteorlake, and zen4

nepnep1111:
The cpu opt patches have support for raptorlake, meteorlake, and zen4 cpus now. GCC13 requirements were mentioned in the customization.cfg

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -188,8 +188,8 @@ _zenify="true"
 _compileroptlevel="1"
 
 # CPU compiler optimizations - Defaults to prompt at kernel config if left empty
-# AMD CPUs : "k8" "k8sse3" "k10" "barcelona" "bobcat" "jaguar" "bulldozer" "piledriver" "steamroller" "excavator" "zen" "zen2" "zen3" (zen3 opt support depends on GCC11)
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Thanks nepnep1111 ! :frog: :heart:

STiNG645:
Trying to build 5.15 with raptorlake. There is no option for raptorlake in menuconfig. I am using GCC 13.2, and I can see from temporary generated .config file that using GCC 13.2. What is the problem?

`# Automatically generated file; DO NOT EDIT.
# Linux/x86 5.15.123 Kernel Configuration
#
CONFIG_CC_VERSION_TEXT="gcc (GCC) 13.2.0"`

nepnep1111:
Just checked the graysky cpu opt repo and it seems that 5.17 is the earliest kernel with raptorlake optimization patches.

STiNG645:
> Just checked the graysky cpu opt repo and it seems that 5.17 is the earliest kernel with raptorlake optimization patches.

Intel native will work correctly, what do you think? Or it's better to build newer versions?

PR #660, opened 2022-12-06
Enable MGLRU support in prjc

ViNi-Arco:
Hey Mr. Etienne, since [mglru is enabled by default](https://github.com/Frogging-Family/linux-tkg/blob/b8835ca20232e4bbf0738bd4e25bcffeb917ad57/customization.cfg#L162) it is worth adding support in BMQ/PDS


Tk-Glitch:
True true! Thanks again ViNi-Arco ! :frog: :heart:

Issue #669, opened 2022-12-21
Kernel won't install on Nobara


Faugus:
Do you want to install the new Kernel ? Y/[n]: Y
Last metadata expiration check: 0:05:03 ago on qua 21 dez 2022 11:09:17.
Error: 
 Problem: problem with installed package kernel-devel-matched-6.0.14-201.fsync.fc36.x86_64
  - package kernel-devel-matched-6.0.13-201.fsync.fc36.x86_64 requires kernel-devel = 6.0.13-201.fsync.fc36, but none of the providers can be installed
  - package kernel-devel-matched-6.0.14-201.fsync.fc36.x86_64 requires kernel-devel = 6.0.14-201.fsync.fc36, but none of the providers can be installed
  - cannot install both kernel-devel-6.0.14_tkg_pds-1.x86_64 and kernel-devel-6.0.13-201.fsync.fc36.x86_64
  - cannot install both kernel-devel-6.0.14_tkg_pds-1.x86_64 and kernel-devel-6.0.14-201.fsync.fc36.x86_64
  - conflicting requests
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
 -> exit cleanup done

KagoraTheGrindcoreAddict:
In your customization.cfg file, is winesync set to true?

I'm pretty sure Nobara ships with a winesync patched kernel and only 1 winesync kernel can be installed at a time so you would have to uninstall the default Nobara kernel (PLEASE DO NOT DO THIS unless you really know what you're doing or even better, how to recover from having no bootable kernels installed)

Personally I stopped using winesync just from the hassle of kernel juggling alone and I have not found a single game I play that benefits from the patch.

Issue #609, opened 2022-10-01
Kernel isnt installed after the install.sh script


Tuerpe:
*Before reporting an issue, please check for relevant keywords in the issues search bar, including closed issues. The solution to your issue might be there already !*

**Describe the bug**
 - Which OS do you encounter the bug on?
 - A clear and concise description of what the bug is.
 - Attach the contents of your logs folder by drag & dropping each file into this message.
**Special note for Arch users: please run `makepkg [args] | tee shell-output.log` for more logging!**

Hey,

its my first time building a kernel so it could be it is just my fault.

After compiling the kernel i got asked if i want to insatll it. I typed yes and after that the script said installation successfull.

I rebooted and didnt saw the kernel. Tried to install it manually (kernel, kernel-devel, kernel-headers) and this was the output:

Last metadata expiration check: 2:51:49 ago on Sat 01 Oct 2022 07:37:21 AM CEST.
Error: 
 Problem: problem with installed package kernel-devel-matched-5.19.11-200.fc36.x86_64
  - package kernel-devel-matched-5.19.11-200.fc36.x86_64 requires kernel-devel = 5.19.11-200.fc36, but none of the providers can be installed
  - package kernel-devel-matched-5.17.5-300.fc36.x86_64 requires kernel-devel = 5.17.5-300.fc36, but none of the providers can be installed
  - cannot install both kernel-devel-5.19.12_tkg_bmqtuerpe-1.x86_64 and kernel-devel-5.19.11-200.fc36.x86_64
  - cannot install both kernel-devel-5.19.12_tkg_bmqtuerpe-1.x86_64 and kernel-devel-5.17.5-300.fc36.x86_64
  - conflicting requests
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)

hostnamectl:
Icon name: computer-desktop
         Chassis: desktop 🖥
      Machine ID: 9a3fe95d51c44bc99c3294b5253bc62e
         Boot ID: 5f397ea47a57448b91eb49d04b3bf033
Operating System: Fedora Linux 36 (Workstation Edition)
     CPE OS Name: cpe:/o:fedoraproject:fedora:36
          Kernel: Linux 5.19.11-200.fc36.x86_64
    Architecture: x86-64
 Hardware Vendor: System manufacturer
  Hardware Model: System Product Name

It would be nice to know what the issue was, but mainly i wrote this for evaluating if it was my fault or if the script should get updated.

[...]

PR #672, opened 2022-12-24
Improve customization of source and work folders

AdelKS:
After our little discussion [here](https://github.com/Frogging-Family/linux-tkg/commit/f4324da8a1f4bad740ac66aeefb652a061666a49), I am opening this PR to improve and unify the `PKGBUILD` and the `install.sh` approach with respect to dealing with kernel sources.

Now:
- We can customize where to put the compressed kernel sources (a git bare repo, size around ~400MB)
- We can customize where to put the "checked out" kernel sources, including a tmpfs folder if wanted.

The cool thing @Tk-Glitch is that you can try to have the compressed kernel sources still in the HDD on your machine and see if it actually improves your compile times, since  it's compressed, and creating a worktree to tmpfs should be faster than cloning ? Let's see !

I need to test on Ubuntu and Fedora to see that I didn't break anything.

### DIFF
--- a/.github/workflows/build-current-kernel-arch.yml
+++ b/.github/workflows/build-current-kernel-arch.yml
@@ -10,6 +10,8 @@ env:
   _processor_opt: "generic"
   PKGDEST: "/tmp/linux-tkg"
   _debugdisable: "true"
+  _kernel_work_folder: /tmp/linux-src-git
+  _kernel_source_folder: /tmp/linux-kernel.git
   # _modprobeddb: "true"
   # _modprobeddb_db_path: ${{ github.workspace }}/modprobed.db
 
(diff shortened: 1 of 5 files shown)


Tk-Glitch:
Looks pretty cool Adel! I'll test the speed but for a more balanced setup than 1gbps fiber + 5400rpm HDD it should definitely be an improvement already.

AdelKS:
Works well on Fedora (I will fix #669 and #609 in another PR) and Ubuntu. I feel like I am missing something but this PR is ready for review and merger.

One merge please -frog v2

Tk-Glitch:
Looks fine to me Adel ! Thanks :heart:
Let's test it against real people :frog:

Issue #679, opened 2023-01-03
0007-v5.15-fsync1_via_futex_waitv.patch fails with kernel 5.15.86


mabod:
This is on EndeavourOS.

with setting `_fsync="true"` the patch `0007-v5.15-fsync1_via_futex_waitv.patch` fails with the following log:

```
### Applying 0007-v5.15-fsync1_via_futex_waitv.patch... ###
patching file include/uapi/linux/futex.h
can't find file to patch at input line 54
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/kernel/futex.c b/kernel/futex.c
|index 4a9e7ce3714a..c3f2e65afab8 100644
|--- a/kernel/futex.c
|+++ b/kernel/futex.c
--------------------------
File to patch: 
Skip this patch? [y] 
Skipping patch.
4 out of 4 hunks ignored
```

Looks like file `kernel/futex.c` does not exist in the source

Iglu47:
Thanks, I totally forgot about that part.

PR #676, opened 2023-01-01
linux 5.15.y: futex_waitv fixup

Iglu47:
following https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v5.15.86&id=fd8a10d44c11f8f53d640db6b97486045b7e0b81

Fixes: https://github.com/Frogging-Family/linux-tkg/issues/675

PR #680, opened 2023-01-03
linux 5.15.y: Update compatibility fix for old Fsync ABI

Iglu47:
addition to https://github.com/Frogging-Family/linux-tkg/pull/676

Fixes https://github.com/Frogging-Family/linux-tkg/issues/679


Tk-Glitch:
Thanks Iglu47, I forgot about it too ❤️ 🐸

PR #684, opened 2023-01-07
Various fixes for the RPM target

AdelKS:
While trying to fix the `mkspec` script within the kernel, I discovered a few things
- Fedora imposes various hardening flags and other bloat that we could avoid, 
- Fedora produces some debug files that conflict with other `tkg` kernels

It took me some time to find documentation to disable all that, but it's done. There also was mistake in the `install.sh` where I was using the wrong work dir path variable, people shouldn't have been able to build for Fedora without giving absolute paths since we merged the work dir change MR.

Adel

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -249,7 +249,7 @@ if [ "$1" = "install" ]; then
       _extra_ver_str="_${_kernel_flavor}"
     fi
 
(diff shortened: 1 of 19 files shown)


Tk-Glitch:
Nice work Adel. Thank you :frog: :heart:

PR #686, opened 2023-01-09
README: Add BORE to alternative schedulers

ptr1337:


Tk-Glitch:
Yup, thanks prt1337! :frog: :heart:

PR #694, opened 2023-01-18
fsync: Rename toggle options

Iglu47:
enough time has passed, so there should be no more problems associated with the transition to using futex_waitv.
this PR should clean up some the fsync things:
- Rename toggle options and update cfg descriptions for
  - futex_waitv backported patches
  - fsync legacy (known as `FUTEX_WAIT_MULTIPLE (opcode 31)`) can still be enabled 
  and used these days for Valve Proton builds (4.11, 5.0 and 5.13)
  - futex2 should be deprecated
- ~update v5.10-fsync_legacy due to [upstream change](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/kernel/futex?h=v5.10.163&id=4750cac4df681ae7bcfa294312f340238b3eadab)~
- remove v5.17-fsync and v5.18-fsync as duplicate files


Iglu47:
probably the only negative is that there is no futex_waitv patch for the 5.10-lts branch

Iglu47:
for some reason renaming and editing a file in the same PR breaks the history of the file.
so I'll leave editing `0007-v5.10-fsync_legacy.patch` for later

Iglu47:
any wishes for this PR? I would like to remove the mention of futex_waitv as a toggle option on the kernel side too.

AdelKS:
@Tk-Glitch Do you have any issues with this PR, it only affects old kernels, so the config option rename should bother anyone :thinking:

Tk-Glitch:
No issue with the PR. Just life stuff slowing some things down. Let's merge. Thank you Iglu47 :frog: :heart:

Iglu47:
Thanks for Your hard work 💛

PR #710, opened 2023-02-22
6.2: Update prjc patchset

ptr1337:
Sync PRJC Scheduler with upstream.
Please consider to enable `CONFIG_PSI_DEFAULT_DISABLED` as default to avoid issues with this patchset.
This issue is still present


Tk-Glitch:
Considering the slow resolution, yes indeed `CONFIG_PSI_DEFAULT_DISABLED` enabled feels like a necessity these days. Thanks ptr1337! :frog: :heart:

Issue #712, opened 2023-03-03
Error applying patches


Andreiva13:
On 6.2.2 I got this error:
`-> Applying Glitched PDS patch`
`==> ERROR: An error was encountered applying patches. It was logged to the prepare.log.txt file.`
[prepare.log.txt](https://github.com/Frogging-Family/linux-tkg/files/10885915/prepare.log.txt)

ptr1337:
Probably introduced by https://github.com/Frogging-Family/linux-tkg/compare/e1372ce04b21...ee6cc74e03bb

Maybe try to revert the commit and check if it works again. The prjc patch needs to be rebased against that, or these needs to be removed.

Andreiva13:
Reverting [ee6cc74e03bb4911d2a7a90be5ccd74bb21dbf4a](https://github.com/Frogging-Family/linux-tkg/commit/ee6cc74e03bb4911d2a7a90be5ccd74bb21dbf4a) fixed this

Rydgel:
Same with Bore

ptr1337:
https://github.com/Frogging-Family/linux-tkg/pull/713

TURING-V2:
applied pr #713 still fail
 * ERROR: sys-kernel/gentoo-sources-6.2.2::gentoo failed (prepare phase):
 *   patch -p1  failed with /etc/portage/patches/sys-kernel/gentoo-sources-6.2.2/0005-glitched-pds.patch

dinckelman:
BORE patches also fail to apply on 6.2.2 with the following [log](https://termbin.com/nlnxr)

ptr1337:
> BORE patches also fail to apply on 6.2.2 with the following [log](https://termbin.com/nlnxr)

I think that should be fixed with the above PR. But actually TKG should revert most of the added patches above, since they hardly conflict with other schedulers, which is not nice.

PR #713, opened 2023-03-04
6.2: Remove fair: Remove all energy efficiency functions

ptr1337:
Fixes https://github.com/Frogging-Family/linux-tkg/issues/712


ventureoo:
Thats not fix the issue, because origin of it is the swappiness parameter, with is already set in:
https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-patches/6.2/0005-glitched-pds.patch#L86
https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-patches/6.2/0003-glitched-base.patch#L1106

ptr1337:
Then I close this. If you want send a new PR. Or TKG will fix this

PR #728, opened 2023-04-24
linux 6.2.y: Replace "Remove all energy efficiency functions" with the default value of sched_energy_aware

ventureoo:
I think that we should not deprive the user of a certain functionality when it is possible to disable it safely without removing the code. Energy efficiency features can be useful for laptop users, so we should leave it up to the user to turn it back on via sysctl.

It also makes rebasing a little easier.


Tk-Glitch:
Thank you ventureoo! :frog: :heart:

PR #747, opened 2023-05-27
Update prepare to include BORE Scheduler

redglobuli:
updated to include BORE scheduler, BORE patch is supporting linux 6.3


Tk-Glitch:
I wanted to do that when going back home. Thanks redglobuli ! :frog: :heart:

Issue #751, opened 2023-05-27
Trying to compile the kernel 6.3.4 with error :((


Everest10:
I'm fedora 38 user and trying to compile the kernel 6.3.4 pds and get this error. how to fix it ?

sh ./scripts/package/mkspec >./kernel.spec
============ WARNING ===========
Your working tree has diff from HEAD, and also untracked file(s). Please make sure you did 'git add' for all new files you need in the source package.
===============================
rpmbuild --define '_topdir /home/dom/linux-tkg//linux-src-git/linux-tkg-rpmbuild' --target x86_64-linux -bs kernel.spec \
--define='_smp_mflags %{nil}' --define='_sourcedir rpmbuild/SOURCES' --define='_srcrpmdir .'
Construindo plataformas de destino: x86_64-linux
Construindo para o destino x86_64-linux
warning: source_date_epoch_from_changelog set but %changelog is missing
Gravou: ./kernel-6.3.4_tkg_pds_g17f2d782f18c-1.src.rpm

RPM build warnings:
    source_date_epoch_from_changelog set but %changelog is missing

Everest10:
Could someone please comment on this issue?

AdelKS:
Hello,

I'm here to comment ! Isn't this issue the same as the one you opened in #738 ?

Everest10:
> Hello,
> 
> I'm here to comment ! Isn't this issue the same as the one you opened in #738 ?
> 
> PS: oh and also #735 thinking Please keep it to one issue if these are all the same problem.
> 
> PPS: please follow the issue template and attach the logs in the 'logs' folder, simply by dragging and dropping them into your message. People then will be able to have a look when possible, if possible.
[customization.cfg.txt](https://github.com/Frogging-Family/linux-tkg/files/11587389/customization.cfg.txt)
[prepare.log.txt](https://github.com/Frogging-Family/linux-tkg/files/11587390/prepare.log.txt)
[shell-output.log.txt](https://github.com/Frogging-Family/linux-tkg/files/11587391/shell-output.log.txt)
[system-info.txt](https://github.com/Frogging-Family/linux-tkg/files/11587392/system-info.txt)

OK. Here is the log files ...

AdelKS:
Okay now we have all the information we need, here's the error that's been happening on your end:
```
make[5]: *** No rule to make target 'kernel/sched/alt_core.o', needed by 'kernel/sched/built-in.a'.  Stop.
[...]

PR #752, opened 2023-06-01
Fix install.sh for fedora

whitehara:
I fixed the issue #751

- Cause
install.sh uses rpm-pkg to build rpms for Fedora.
The rpm-pkg makes diff.patch from the patched source tree, pack it into the src.rpm, then extract vanilla source and apply diff.patch.
For making diff.patch, "git diff HEAD" is used. But "git diff HEAD" can not detect modified files without "git add".
So, new files from patches are not created after applying the patch during rpm-pkg build.

- Solution
I added "git add" into the install.sh for detecting the patched files before rpm-pkg.
(I also added "dnf install perl" for the build dependency)


Everest10:
Now I can compile the 6.3.5 pds for fedora 38.

whitehara:
> Have you checked that it works well until the end ?
Yes, I have checked on Fedora 38 for compiling 6.3.5_tkg_bmq, and it finished building the rpm.

> Can we still rebuild another kernel just fine ? 
I didn't check other case for Debian, etc.

PR #773, opened 2023-07-24
Add bcachefs for 6.5

obj-obj:
Currently untested, this will remain a draft PR until I make sure nothing goes wrong


ptr1337:
These patches are completly malformed, besides the bcachefs one from lucjan

obj-obj:
> These patches are completly malformed, besides the bcachefs one from lucjan

Yeah, I'll have to fix it tomorrow

I got them directly from the linux-surface github, but I think something went wrong when I downloaded them

simifor:
Haven't done a lot of testing, but bcachefs seems to work fine, I can read and write files as expected. I only grabbed that patch so I can't say anything about the other patches in this PR

obj-obj:
Oh, it was already added anyway

PR #799, opened 2023-08-31
Remove anbox option

kylon:
officially no longer active since months.
basically no longer active since ages.

an alternative is available and in active development (also recommended by anbox), no need to keep this here


Tk-Glitch:
I'm usually against plain removal when it's still usable/used by some users, but in this case I think you're right and we can get rid of it.
Thanks ! :frog: :heart:

PR #801, opened 2023-08-31
Update user-patches and hardened config docs (closes #779)

### LINKED ISSUE #779: `README.md` and `customization.cfg`: Only kernel version 5.x supported, not 6.x? -- Probably documentation is out of date.
Ahoj,

I am currently making myself familiar with the `linux-tkg` build system.

I have cloned https://github.com/Frogging-Family/linux-tkg.git, and now I read the bundled [`README.md`](/Frogging-Family/linux-tkg/blob/master/README.md).

There it states stuff specific only for 5.x kernels -- but kernels are now already at 6.x:

> #### *User patches*
> 
> *To apply your own patch files using the provided scripts, you will need to put them in a `linux5y-tkg-userpatches` folder -- `y` needs to be changed with the kernel version the patch works on [...]*

Taking the [`README.md`](/Frogging-Family/linux-tkg/blob/master/README.md) verbally, I would assume that for 6.x kernels the userpatches feature is not supported.

If that is not the case, and only the documentation is outdated, then I suggest to re-phrase the documentation to be generic with regard to kernel version -- e.g.:

> #### *User patches*
> 
> *To apply your own patch files using the provided scripts, you will need to put them in a `linuxXY-tkg-userpatches` folder -- `X` and `Y` needs to be changed with the kernel version the patch works on [...]*

Also, the [`customization.cfg`](/Frogging-Family/linux-tkg/blob/master/customization.cfg) has wrong information.

It says  
```
# "config_hardened.x86_64" : config file for a hardened kernel, available for kernel version "5.13", "5.10" and "5.4" .
```

But a `ls */*/*hardened*` reveals that it must be `"5.15"`, `"5.13"`, `"5.11"`, `"5.10"`, `"5.7"` and `"5.4"`.

Also, it seems to support only 5.x kernels out of the box:

```
# Kernel Version - Options are "5.4", and from "5.7" to "5.19"
```

There might be other places where information seems inconsistent with reality.

Regards!

kylon:


Tk-Glitch:
:frog: :heart:

Issue #404, opened 2022-01-18
Resolution changed


RoyleHerman:
After downloading linux-tkg my resoultion changed to 1024x768 and i can't change it.
Dist: Ubuntu 20.04
Gpu: gtx1660
Cpu: i5-9400f

Tk-Glitch:
Make sure you have dkms nvidia drivers installed and that the version used (495.46 should be okay) is compatible with kernel 5.16.

PR #805, opened 2023-09-04
Add _kernel_on_diet customization option

AdelKS:
Hello,

2023 is the year linux-tkg decided to be more wary of its weight and start exercising.

So, this PR adds `_kernel_on_diet` option, that makes use of a bundled-in `minimal-modprobed.db` file to build the kernel. I think the db is still not as big as it should be, but we will keep growing it with everyone's contribution of their db. Then we can count on issues/PRs to add more modules to the list.

This change is motivated by the need to fix github ci built kernels, which stopped to successfully build after kernels got too fat or github lowered the specs of their runners, who knows. I activated this setting in the ci on the master branch of my fork, let's see if it successfully builds.


Tk-Glitch:
Looks like it worked !
#404 for the win !
Thanks Adel, as always <3

ryanmusante:
Interesting, what happens when you compile with a smaller kernel and it needs additional modules? Will it just error out on the ones that were extricated?

Tk-Glitch:
Well, the module won't exist in the package so the app or hardware depending on it will not work. The idea is to make this grow to cover 99% of users needs.

ryanmusante:
OK, I'll build with modprobed.db next time and contribute. I'm running on a fairly typical laptop install.

ryanmusante:
[modprobed.db.txt](https://github.com/Frogging-Family/linux-tkg/files/12554801/modprobed.db.txt)

PR #814, opened 2023-09-15
Add some more kernel modules to diet config

FintasticMan:
Partly taken from my database, partly from @ryanmusante's. Also update the sorting to match what modprobed-db outputs, so that it's easier to check which modules are missing from the diet config.


ryanmusante:
Woot woot

linux-tkg-config/6.5/minimal-modprobed.db
```diff
@@ -109,28 +109,29 @@ cuse
...
 dm_mod
 dm_multipath
 dm_region_hash
 dm_round_robin
+dmi_sysfs
 dns_resolver
 drm
 drm_buddy
 drm_display_helper
+drm_dp_helper
```
AdelKS, line 123: If I remember well, this module is not recognized when actually building the kernel, try to build the kernel first and look for the log messages it outputs about modules that are not recognized. These don't need to be there.

linux-tkg-config/6.5/minimal-modprobed.db
```diff
@@ -383,6 +389,7 @@ nls_utf8
 nouveau
 ntfs
 ntfs3
+nvidia_wmi_ec_backlight
```
AdelKS, line 392: This one is also not in-tree.

ryanmusante:
image

Those both show up from linux-tkg
~ > lsmod | grep nvidia           
nvidia_wmi_ec_backlight    12288  0
nvidia_uvm           3497984  0
nvidia_drm             94208  5
nvidia_modeset       1556480  8 nvidia_drm
nvidia              62713856  340 nvidia_uvm,nvidia_modeset
video                  77824  2 nvidia_wmi_ec_backlight,nvidia_modeset
wmi                    45056  3 video,nvidia_wmi_ec_backlight,mxm_wmi
~ > pkgfile nvidia_wmi_ec_backlight   
~ > locate drm_dp_helper          
/usr/lib/modules/6.5.3-273-tkg-eevdf/build/include/drm/display/drm_dp_helper.h
~ > pkgfile drm_dp_helper.h         
core/linux419-headers
core/linux510-headers
core/linux515-headers
core/linux54-headers
core/linux61-headers
core/linux61-rt-headers
core/linux64-headers
core/linux64-rt-headers
core/linux65-headers
core/linux65-rt-headers
core/linux66-headers

linux-tkg-config/6.5/minimal-modprobed.db
FintasticMan: Sorry for taking so long to respond. Indeed, I do get a warning saying `drm_dp_helper config not found!!`. I'll remove that one then.

FintasticMan: Hmm, I've just noticed that it does also build the module though... `CC [M]  drivers/gpu/drm/display/drm_dp_helper.o`

linux-tkg-config/6.5/minimal-modprobed.db
FintasticMan: I don't get any warnings about this module, and it does build a `nvidia-wmi-ec-backlight.ko` file, so I think it is in-tree.

linux-tkg-config/6.5/minimal-modprobed.db
```diff
@@ -352,16 +357,17 @@ nf_nat_irc
...
 nfnetlink_log
 nfnetlink_queue
-nf_reject_ipv4
-nf_reject_ipv6
 nfs
 nfsv4
-nf_tables
 nft_chain_nat
 nft_compat
+nft_counter
```
FintasticMan, line 370: This module also doesn't seem to be in-tree, as I get a log about `nft_counter config not found!!`. I'll remove this one too then.

linux-tkg-config/6.5/minimal-modprobed.db
AdelKS: Then I probably thought it wasn't there just by reading `nvidia` haha. All good then.

linux-tkg-config/6.5/minimal-modprobed.db
```diff
@@ -25,9 +25,9 @@ algif_hash
 algif_skcipher
 alx
 amd64_edac
-amdgpu
```
AdelKS, line 28: Why  is the order changing? Are you putting this list in `~/.config/modprobed.db` then calling `modprobed-db store` to get its own ordering ?

FintasticMan: I looked at the modprobed-db script and checked how it sorts, then applied that to the file. It uses `sort -k 1,1` to sort the file.

AdelKS: I just did the procedure I suggested on `minimal-modprobed.db` of this PR and the order does not get changed with respect to master.

[minimal-modprobed.db.txt](https://github.com/Frogging-Family/linux-tkg/files/12840986/minimal-modprobed.db.txt)

So probably the tool `modprobed-db` is doing something more. If you can figure out exactly how it's ordered would be nice yeah. Otherwise just run modprobed-db on it and we're done x)

FintasticMan: I've just run `modprobed-db store` on the `minimal-modprobed.db` from master, and it changes the ordering to the one in this PR, so that combined with reading the code makes me quite sure that `sort -k 1,1` is the correct ordering.

FintasticMan:
I've just added some more modules from my laptop's modprobed-db.

linux-tkg-config/6.5/minimal-modprobed.db
AdelKS: Ah that's interesting, I have no idea why my machine gives the other order. `sort` comes from the `coreutils` package so it doesn't depend on the shell :thinking:

FintasticMan: It might be that it doesn't sort it if there are no modules to be added?

Tk-Glitch:
Thanks @FintasticMan and @AdelKS 
Merging ! :frog: :heart:

linux-tkg-config/6.5/minimal-modprobed.db
AdelKS: Just figured it out, I need `LC_ALL=C` to get the same sorting as you

Issue #817, opened 2023-09-19
bore-eevdf: kernel panic on 6.5.4


v1993:
**Describe the bug**
Running on up-to-date Manjaro, I ran into a kernel panic while trying to build a kernel with different options (namely LLVM and Thin LTO). I'm not sure when exactly this problem started: I've used 6.5.3 PDS to build the problematic 6.5.4 bore-eevdf kernel, although at least 6.5.2 bore-eevdf used to build *that* apparently did not exhibit this problem (then again, this is the first time I'm building kernel using LLVM - but keep in mind that *panicking kernel* was built with GCC).

Kernel panic screen:

![photo of kernel panic screen](https://github.com/Frogging-Family/linux-tkg/assets/16255906/7ce17071-b929-42b4-bcb6-d69e1c9f5c0f)

Contents of `/usr/share/doc/linux65-tkg-bore-eevdf/customization.cfg`:

```sh
_CE_CONDA=""
_CE_M=""
_N="5"
_NR_CPUS_value=""
_NUKR="true"
_STRIP="true"
_acs_override=""
_aggressive_ondemand="true"
_alg="reno"
_alg_upper="RENO"
_avail_cpu_scheds=([0]="cfs"
_avail_cpu_scheds_text=([0]="CFS
_avail_timer_frequencies=([0]="100"
_avail_timer_frequencies_text=([0]="100
_basekernel="6.5"
_basever="65"
_bcachefs="false"
_cacule_rdb="false"
_cacule_rdb_interval="19"
_clear_patches="true"
_community_patches=""
_compiler="gcc"
_compiler_name=""
_compileroptlevel="1"
_config_fragments="true"
_config_fragments_no_confirm="false"
_config_name="VIRT_CPU_ACCOUNTING_GEN"
_config_updating="olddefconfig"
_configfile=""
_cpu_marchs=([0]="native_amd"
_cpusched="bore-eevdf"
_cur_char="4"
_cur_index="2"
_current_kernels=([0]="6.6"
_custom_commandline="intel_pstate=passive
_custom_pkgbase=""
_debugdisable="false"
_default_cpu_gov="ondemand"
_deprecated_config_var_set=([_use_tmpfs]="1"
_deprectated_config_var="_source_in_tmpfs"
_diffconfig=""
_diffconfig_name=""
_distro="Arch"
_dracut_options="--lz4"
_eol_kernels=([0]="6.3"
_f="/home/v/compile/linux-tkg/*.myrevert"
[...]

Issue #819, opened 2023-09-22
freezes with 6.5.4-273.1-tkg-eevdf


mabod:
I wanted to try 6.5.4 kernel with eevdf. I booted the kernel and started a kernel compilation to check how the PC behaves under load. But it froze after a few minutes. And it could not even be rebooted with REISUB.

The journal shows a warning/error right after boot:

```
Sep 22 10:44:17 rakete kernel: Linux version 6.5.4-273.1-tkg-eevdf (linux65-tkg-eevdf@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0) #1 SMP PREEMPT_DYNAMIC TKG Fri, 22 Sep 2023 05:08:26 +0000
[...]
Sep 22 10:44:17 rakete kernel: psi: inconsistent task state! task=125:cpuhp/18 cpu=18 psi_flags=4 clear=0 set=4
```

Later on, at the end of the journal I see this 

```
Sep 22 11:05:31 rakete rtkit-daemon[5577]: The canary thread is apparently starving. Taking action.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Demoting known real-time threads.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 21902 of process 21678.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11509 of process 11488.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11488 of process 11488.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11506 of process 11489.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11489 of process 11489.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11508 of process 11487.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Successfully demoted thread 11487 of process 11487.
Sep 22 11:05:31 rakete rtkit-daemon[5577]: Demoted 7 threads.
Sep 22 11:05:41 rakete rtkit-daemon[5577]: The canary thread is apparently starving. Taking action.
Sep 22 11:05:41 rakete rtkit-daemon[5577]: Demoting known real-time threads.
Sep 22 11:05:41 rakete rtkit-daemon[5577]: Successfully demoted thread 21902 of process 21678.
[...]

PR #820, opened 2023-09-22
Fixup (bore) EEVDF issue in #819

ptr1337:
Don't use the latest EEVDF Patches from peterz/queue. Seems to introduce issues for some configurations.
Fixes: https://github.com/Frogging-Family/linux-tkg/issues/817


Tk-Glitch:
Thanks Peter! I was contemplating a similar solution.
:frog: :heart:

PR #827, opened 2023-10-17
6.5: Add EEVDF urgent fix

sirlucjan:
See: https://git.kernel.org/pub/scm/linux/kernel/git/peterz/queue.git/commit/?h=sched/urgent


Tk-Glitch:
Thanks sirlucjan! That explains some reports.

sirlucjan:
You're welcome.

PR #842, opened 2023-11-11
prepare: llvm: add new options to avoid prompt

kylon:


linux-tkg-config/prepare
```diff
@@ -981,7 +981,11 @@ _tkg_srcprep() {
   fi
 
   if [ "$_compiler_name" = "-llvm" ]; then
+    _enable "CFI_CLANG"
```
AdelKS, line 984: I'd put that one to its default value, disabled. People can then enable it with frags if needed

kylon: ok, thought enabled was default, my bad

Tk-Glitch:
Thanks kylon :)

PR #843, opened 2023-11-11
customization.cfg: modprobeddb_path: add note about chroot

kylon:
i build in chroot, but something seems wrong with this
you set a path, then prepare script prepend $_where to the path, and that should give you an invalid path?
also afaik when executed it will not expand `~`, so the example path should be a wrong example

am i missing something?

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -63,7 +63,8 @@ _kernel_on_diet="false"
 _modprobeddb="false"
 
 # modprobed-db database file location
-_modprobeddb_db_path=~/.config/modprobed.db
+# if building on a chroot, set this to modprobed.db and manually copy the modprobed.db file from ~/.config/modprobed.db to linux-tkg root
+_modprobeddb_db_path="~/.config/modprobed.db"
 
 # Set to "1" to call make menuconfig, "2" to call make nconfig, "3" to call make xconfig, before building the kernel. Set to false to disable and skip the prompt.
 _menunconfig=""


AdelKS:
> you set a path, then prepare script prepend $_where to the path, and that should give you an invalid path?

I went to take a look at the code:

```bash
 if [ -f "$_where"/"$_modprobeddb_db_path" ]; then
      _modprobeddb_db_path="$_where"/"$_modprobeddb_db_path"
    fi
    if [ ! -f $_modprobeddb_db_path ]; then
      msg2 "modprobed-db database not found"
      exit 1
    fi
```

if the path `"$_where"/"$_modprobeddb_db_path"` exists, it takes that, otherwise it takes the path verbatim and checks if it exists. I don't see why this wouldn't work on a chroot :thinking:

kylon:
because the path is not valid and also my home dir is not available in chroot (using arch chroot on the fly with a script)

if i set /home/xxx/.config/modprobed.db or ~/.config/modprobed.db, the results are build/home/xxx/.config/modprobed.db and build/.config/modprobed.db respectively

these should be invalid even outside of chroot

AdelKS:
If you are using a chroot, it should work if you put the absolute path in the config, the absolute path after chrooting, or a relative path with respect to where the PKGBUILD is within the chroot.

kylon:
ok the path was borked with the various branches i had, but still, using [mkchroot](https://wiki.archlinux.org/title/DeveloperWiki:Building_in_a_clean_chroot#Setting_up_a_chroot), home is not available (and i think with chroot in general) so the only option is to either put it in linux-tkg root folder or copy it when you build the chroot or after in a folder available in chroot
i actually never had the need to use home files on chroot, that was my confusion

the same applies for the tkg custom config in home if you are using chroots

AdelKS:
Okay, yeah all of that is to be expected. I don't think there's anything to add in the doc, as the user using the chroot should be aware of these kind of shenanigans :D

PR #844, opened 2023-11-11
Docs, prepare: EEVDFfy

kylon:
not sure what to do with preemprt
~i did not add glitched addition to eevdf, not sure if that is still valid/updated for eevdf~ added eevdf additions patch


linux-tkg-config/prepare
```diff
@@ -252,14 +252,14 @@ _set_cpu_scheduler() {
   _sched_description_array=(
     ["pds"]="Project C / PDS"
     ["bmq"]="Project C / BMQ"
-    ["cfs"]="CFS (Completely Fair Scheduler) Linux kernel's default"
+    ["eevdf"]="EEVDF (Earliest Eligible Virtual Deadline First scheduler) Linux kernel's default [6.6+]"
```
AdelKS, line 255: Probably comparisons are more explicit, you can use UTF chracters `≥ 6.6`

linux-tkg-config/prepare
```diff
@@ -252,14 +252,14 @@ _set_cpu_scheduler() {
...
+    ["cfs"]="CFS (Completely Fair Scheduler) Linux kernel's default [6.5-]"
```
AdelKS, line 256: Same here `≤ 6.5`, especially because the dash can mean `6.5-6.9` or something

linux-tkg-config/prepare
AdelKS:
```suggestion
    ["eevdf"]="EEVDF (Earliest Eligible Virtual Deadline First scheduler) Linux kernel's default for ≥ 6.6"
```

linux-tkg-config/prepare
AdelKS:
```suggestion
    ["cfs"]="CFS (Completely Fair Scheduler) Linux kernel's default for ≤ 6.5"
```

README.md
```diff
@@ -15,21 +15,21 @@ This repository provides scripts to automatically download, patch and compile th
 ### Customization options
 #### Alternative CPU schedulers
 
-[CFS](https://en.wikipedia.org/wiki/Completely_Fair_Scheduler) is the only CPU scheduler available in the "vanilla" kernel sources. Its current implementation doesn't allow for injecting additional schedulers, and requires replacing it. Only one scheduler can be patched in at a time.
+[EEVDF](https://lwn.net/Articles/925371/) is the only CPU scheduler available in the "vanilla" kernel sources. Its current implementation doesn't allow for injecting additional schedulers, and requires replacing it. Only one scheduler can be patched in at a time.
```
AdelKS, line 18: Probably best to keep both and say which versions of the kernel just like with the `prepare` function

linux-tkg-config/prepare
```diff
@@ -304,7 +304,7 @@ _set_cpu_scheduler() {
   elif [ "$_kver" = "606" ]; then
     _avail_cpu_scheds=("eevdf" "bore")
   else
-    _avail_cpu_scheds=("cfs")
+    _avail_cpu_scheds=("cfs" "eevdf")
```
AdelKS, line 307: actually here, if I remember well, this block for the newest rc kernels, so you can simply put `eevdf` alone.

linux-tkg-config/prepare
```diff
@@ -348,7 +348,7 @@ _set_cpu_scheduler() {
   if [ -z "$_cpusched" ]; then
     msg2 "Which CPU sched variant do you want to build/install?"
     msg2 "Project C (pds) / BMQ (bmq) is usually a good balance for gaming."
-    msg2 "Select \"cfs\" (linux kernel's default) if unsure."
+    msg2 "Select \"eevdf\" (linux kernel's default) if unsure."
```
AdelKS, line 351: I think we should define a `_default_scheduler` variable and populate it when the kernel version is chosen (`cfs` or `eevdf`). then just use it everywhere where it makes sense, like here.

AdelKS (changes requested):
Thanks a bunch for putting in the efforts ! There are some things we can improve a little bit :D

linux-tkg-config/prepare
```diff
@@ -6,6 +6,9 @@ _current_kernels=("6.6" "6.5" "6.4" "6.1" "5.15")
 # List of kernels that are no longer maintained either upstream or locally
 _eol_kernels=("6.3" "6.2" "6.0" "5.19" "5.18" "5.17" "5.16" "5.14" "5.13" "5.12" "5.11" "5.9" "5.8" "5.7" "5.4.230" "5.10.135")
 
+# Current CPU scheduler in vanilla Linux kernel
+_default_cpu_sched="eevdf"
```
AdelKS, line 10: Great thanks! Sorry for thinking about it just now: 

it's best to define `_default_cpu_sched` at the end of `_set_kver_internal_vars()`: this function is called right after the kernel is chosen, to define some internal vars. And we can use `_kver` to know which scheduler is the default:

```bash

if [ $_kver -le 605 ]; then
  # chosen kernel <= 6.5
  _default_cpu_sched="cfs"
else
  # chosen kernel >= 6.6
  _default_cpu_sched="eevdf"
fi

 echo -e "_basekernel='$_basekernel'\n_basever='$_basever'\n_sub='$_sub'\n_kver='$_kver'\n_default_cpu_sched='$_default_cpu_sched'" > "$_where"/BIG_UGLY_FROGMINER
```

AdelKS:
@Tk-Glitch I will leave reviewing the patches to you, I never really followed that x)

linux-tkg-config/prepare
```diff
@@ -179,6 +179,14 @@ _set_kver_internal_vars() {
...
 
+  if [ $_kver -le 605 ]; then
+    # chosen kernel <= 6.5
+    _default_cpu_sched="cfs"
+  else
+    # chosen kernel >= 6.6
+    _default_cpu_sched="eevdf"
+  fi
+
   echo -e "_basekernel='$_basekernel'\n_basever='$_basever'\n_sub='$_sub'\n_kver='$_kver'" > "$_where"/BIG_UGLY_FROGMINER
```
AdelKS, line 190: I think we need to also save it in the file `BIG_UGLY_FROGMINER` too, so it's not lost when contexts change
```suggestion
# add it the the file BIG_UGLY_FROGMINER (because the bash script is sometimes called from different contexts where the global variables get reset, so this file is sourced every time)
 echo -e "_basekernel='$_basekernel'\n_basever='$_basever'\n_sub='$_sub'\n_kver='$_kver'\n_default_cpu_sched='$_default_cpu_sched'" > "$_where"/BIG_UGLY_FROGMINER
```

Tk-Glitch:
The glitched-cfs-additions patch will not work with eevdf as is (the fair.c bits at least). Otherwise it looks fine at first glance.

kylon:
what do you mean? i removed that, the new one is eevdf-glitched for 6.6

Tk-Glitch:
Ok, looks fine to me. Thanks kylon!
Merging.
:frog: :heart:

PR #845, opened 2023-11-12
install.sh: fix handling of env var config overrides

AdelKS:
overriding config values through environment variables wasn't working because the script call was resetting them before
they were saved

also, the external config file was getting sourced twice


Tk-Glitch:
Nice. Thanks Adel! :frog: :heart:

PR #848, opened 2023-11-15
In case of Fedora: modify kernel configuration to add SELinux. Fix #383

### LINKED ISSUE #383: [Fedora 35] No longer able to boot Fedora's stock kernel
Before I installed linux-tkg, I was able to boot the stock kernel just fine. But since then, every time I boot up the stock kernel, the splash screen spits out something like this:
```
[FAILED] Failed to start D-Bus System Message Bus. //repeated for like 6 more times
[FAILED] Failed to start Bluetooth service.
//complains about alsactl
systemd-shutdown
```
Despite all that, linux-tkg still boots up just fine but I'd like to be able to use stock kernel just in case. 
Perhaps the fact that it replaces the stock kernel-headers and kernel-devel is related to this?

Chevek:
Add Fedora kernel configuration for SELinux. This prevent relabel of most files when booting a kernel without this feature.

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1781,6 +1781,16 @@ CONFIG_DEBUG_INFO_BTF_MODULES=y\r
   msg2 "Setting config"
   make ${_config_updating} ${llvm_opt} |& tee -a "$_where"/logs/prepare.log.txt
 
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thanks Chevek, looks fine to me. Fedora users will have a smoother run now.
:frog: :heart:

PR #849, opened 2023-11-15
README: remove Fedora warning

AdelKS:
now that we have fixed selinux


Tk-Glitch:
:+1: Thanks Adel! :frog: :heart:

PR #857, opened 2023-12-08
6.6: Add patches for WRITE_WATCH support in Wine

kakra:
This adds missing write watch patches I've ported over from 6.1 to 6.6. I'm using this in my own daily-drive 6.6 LTS kernel.

See-also: https://github.com/kakra/linux/pull/30


Tk-Glitch:
Thanks kakra! :frog: :heart:

PR #860, opened 2023-12-22
Fix: Dracut is not being run on Gentoo

Slaviusz:
Since refactorings initramfs was not being built if distro was set to "Gentoo".

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -411,14 +411,12 @@ if [ "$1" = "install" ]; then
         sudo emerge @module-rebuild --keep-going
       fi
 
-    else
-
-      msg2 "Creating initramfs"
-      sudo dracut --force --hostonly ${_dracut_options} --kver $_kernelname
-      msg2 "Updating GRUB"
-      sudo grub-mkconfig -o /boot/grub/grub.cfg
-
     fi
+
+    msg2 "Creating initramfs"
+    sudo dracut --force --hostonly ${_dracut_options} --kver $_kernelname
+    msg2 "Updating GRUB"
+    sudo grub-mkconfig -o /boot/grub/grub.cfg
 
   fi
 fi


AdelKS:
Actually I removed the dracut call from Gentoo code path because Gentoo has its own `kernel-install` script that calls dracut. If it doesn't happen with your machine probably there's some kind of configuration to do. I cannot find documentation for now.

Slaviusz:
On my system the dracut does not run when `./install sh install` is executed. I will look into the official gentoo kernel ebuilds to see how they trigger dracut or any other ramfs tool and will update this PR. Please keep it open. Thanks and Merry Christmas.

AdelKS:
Ah I see why know, the wrapper script `kernel-install`, that is called when doing `make install` belongs to the `systemd` package. Let me ask around in the IRC

AdelKS:
And since not everyone uses `dracut` nor `grub`, it wasn't very good to force that code

AdelKS:
Okay probably to a recent change in Gentoo, I just read an eselect news that redirects to here: https://wiki.gentoo.org/wiki/Installkernel

So I think the current approach of not calling `dracut` is correct

ThisNekoGuy:
Wouldn't the correct approach be to check if it's in path first rather than just not calling it?

AdelKS:
What do you want check in the path ? 

[Installkernel](https://wiki.gentoo.org/wiki/Installkernel) currently handles `grub`, `systemd-boot` and `dracut`. Probably with a little bit more time more modules will be added to support other tools. That's way better than adding extra code paths in this repository which add maintenance.

ThisNekoGuy:
Doesn't install-kernel just trigger whenever portage manages kernels? I've had to manually include dracut in the script because of this issue, so I don't see what you mean about it doing its job.

AdelKS:
The package [sys-kernel/installkernel](https://packages.gentoo.org/packages/sys-kernel/installkernel) overrides the script that gets called when you do `make install` after building the linux kernel by hand, and our `install.sh` script does call `make install`.  I can assume that you are not using `systemd` either, have you installed that package with `-systemd +dracut` ?

ThisNekoGuy:
> I can assume that you are not using `systemd` either, have you installed that package with `-systemd +dracut`?

Yeah

AdelKS:
Look at the documentation here: https://github.com/gregkh/linux/blob/master/arch/x86/Makefile#L330
`make install` is supposed to call `/bin/$(INSTALLKERNEL)`

ThisNekoGuy:
This is the only line even mentioning `dracut` in `installkernel`:
```bash
# use the same input path as /usr/lib/kernel/install.d/50-dracut.install
# and the same output path as dracut and genkernel-4 for best interop
base_dir=$(dirname "${img}")
initrd=${base_dir}/initrd
uki=${base_dir}/uki.efi
```

AdelKS:
This is the hook that is supposed to be installed https://github.com/projg2/installkernel-gentoo/blob/master/hooks/50-dracut.install

ThisNekoGuy:
I have the script there, and correct me if I'm wrong, but it isn't completely clear to me looking at it:

<details><summary>File:</summary>

```bash
#!/bin/bash

COMMAND="$1"
KERNEL_VERSION="$2"
BOOT_DIR_ABS="$3"
KERNEL_IMAGE="$4"

# If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
# So, let's skip to create initrd.
if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
    exit 0
fi

# Do not attempt to create initramfs if the supplied image is already a UKI
if [[ "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
    exit 0
fi

# Mismatching the install layout and the --uefi/--no-uefi opts just creates a mess.
if [[ $KERNEL_INSTALL_LAYOUT == "uki" && -n $KERNEL_INSTALL_STAGING_AREA ]]; then
    BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"
    if [[ -z $KERNEL_INSTALL_UKI_GENERATOR || $KERNEL_INSTALL_UKI_GENERATOR == "dracut" ]]; then
        # No uki generator preference set or we have been chosen
        IMAGE="uki.efi"
        UEFI_OPTS="--uefi"
    elif [[ -z $KERNEL_INSTALL_INITRD_GENERATOR || $KERNEL_INSTALL_INITRD_GENERATOR == "dracut" ]]; then
        # We aren't the uki generator, but we have been requested to make the initrd
        IMAGE="initrd"
        UEFI_OPTS="--no-uefi"
    else
        exit 0
    fi
elif [[ $KERNEL_INSTALL_LAYOUT == "bls" && -n $KERNEL_INSTALL_STAGING_AREA ]]; then
    BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"
    if [[ -z $KERNEL_INSTALL_INITRD_GENERATOR || $KERNEL_INSTALL_INITRD_GENERATOR == "dracut" ]]; then
        IMAGE="initrd"
        UEFI_OPTS="--no-uefi"
    else
        exit 0
    fi
else
    # No layout information, use users --uefi/--no-uefi preference
    UEFI_OPTS=""
    if [[ -d $BOOT_DIR_ABS ]]; then
        IMAGE="initrd"
    else
        BOOT_DIR_ABS="/boot"
        IMAGE="initramfs-${KERNEL_VERSION}.img"
    fi
fi

ret=0

case "$COMMAND" in
    add)
        if [[ $IMAGE == "uki.efi" ]]; then
            IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/uki.efi
        else
            IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/initrd
        fi
        if [[ -f ${IMAGE_PREGENERATED} ]]; then
            # we found an initrd or uki.efi at the same place as the kernel
            # use this and don't generate a new one
            [[ $KERNEL_INSTALL_VERBOSE == 1 ]] && echo \
                "There is an ${IMAGE} image at the same place as the kernel, skipping generating a new one"
            cp --reflink=auto "$IMAGE_PREGENERATED" "$BOOT_DIR_ABS/$IMAGE" \
                && chown root:root "$BOOT_DIR_ABS/$IMAGE" \
                && chmod 0600 "$BOOT_DIR_ABS/$IMAGE" \
                && exit 0
        fi

        if [ -n "$KERNEL_INSTALL_CONF_ROOT" ]; then
            if [ -f "$KERNEL_INSTALL_CONF_ROOT/cmdline" ]; then
                read -r -d '' -a BOOT_OPTIONS < "$KERNEL_INSTALL_CONF_ROOT/cmdline"
            fi
        elif [[ -f /etc/kernel/cmdline ]]; then
            read -r -d '' -a BOOT_OPTIONS < /etc/kernel/cmdline
        elif [[ -f /usr/lib/kernel/cmdline ]]; then
            read -r -d '' -a BOOT_OPTIONS < /usr/lib/kernel/cmdline
        else
            declare -a BOOT_OPTIONS

            read -r -d '' -a line < /proc/cmdline
            for i in "${line[@]}"; do
                [[ ${i#initrd=*} != "$i" ]] && continue
                BOOT_OPTIONS+=("$i")
            done
        fi

        unset noimageifnotneeded

        for ((i = 0; i < "${#BOOT_OPTIONS[@]}"; i++)); do
            # shellcheck disable=SC1001
            if [[ ${BOOT_OPTIONS[$i]} == root\=PARTUUID\=* ]]; then
                noimageifnotneeded="yes"
                break
            fi
        done

        # shellcheck disable=SC2046
        dracut -f \
            ${noimageifnotneeded:+--noimageifnotneeded} \
            $([[ $KERNEL_INSTALL_VERBOSE == 1 ]] && echo --verbose) \
            $([[ -n $KERNEL_IMAGE ]] && echo --kernel-image "$KERNEL_IMAGE") \
            "$UEFI_OPTS" \
            --kver "$KERNEL_VERSION" \
            "$BOOT_DIR_ABS/$IMAGE"
        ret=$?
        ;;

    remove)
        rm -f -- "$BOOT_DIR_ABS/$IMAGE"
        ret=$?
        ;;
esac

exit $ret
```
</details>

AdelKS:
The question is whether or not that hook is getting triggered, it probably isn't. Do you have an `installkernel` in your path ? it should be installed by `sys-kernel/installkernel` and `make install` should call it, afaik :thinking:

ThisNekoGuy:
`installkernel` is in my path, in `/sbin`, yeah.
I'll omit my inclusion of dracut in the script soon, when I have a moment, and try again to see if it triggers but I do remember needing to have put it there a while ago because my `/boot` folder didn't have the proper files after I ran the script one day

AdelKS:
Okay I think I have now the final answer to this: install `sys-kernel/installkernel`, and add `+dracut + grub` useflag there. And that's all you should need.

AdelKS:
This PR is not needed any longer, thanks for bringing it up ! Final solution in #883

PR #861, opened 2023-12-22
Use --ignore-whitespace flag when patching the work tree

Sid127:
Allows patches to apply successfully even with mismatching whitespace characters, making life less hair-pully for people adding their own patches.


openglfreak:
If this gets merged, I'm reverting it locally, because I want all the patches I carry to apply properly and to continue to conform to surrounding coding style.

But if that's not a concern for anyone else then feel free to merge this.

Tk-Glitch:
I'll add a config switch for this so conformance isn't an issue as openglfreak pointed out. But yeah, will merge.

PR #864, opened 2023-12-29
Extra build pipelines for Fedora and Ubuntu

AdelKS:
Hello !

I am back with some extra juicy stuff as usual: build pipelines for DEBs and RPMs. It works: https://github.com/AdelKS/linux-tkg/releases/tag/v6.6.8

I also factorized the pipeline code since github actions added this: https://docs.github.com/en/actions/using-workflows/reusing-workflows (I was waiting for them to add it)

One question, you did this for Arch
```
'CONFIG_FAT_FS=m\nCONFIG_MSDOS_FS=m\nCONFIG_VFAT_FS=m\nCONFIG_FAT_DEFAULT_CODEPAGE=437\nCONFIG_FAT_DEFAULT_IOCHARSET="ascii"\nCONFIG_FAT_DEFAULT_UTF8=y\nCONFIG_EXFAT_FS=m\nCONFIG_EXFAT_DEFAULT_IOCHARSET="utf8"\n# CONFIG_NTFS_FS is not set\nCONFIG_NTFS3_FS=m\n# CONFIG_NTFS3_64BIT_CLUSTER is not set\nCONFIG_NTFS3_LZX_XPRESS=y\nCONFIG_NTFS3_FS_POSIX_ACL=y' > FS_modprobed-db.myfrag
```

I suppose all of that is taken care with the `diet` kernel modprobed-db ? Otherwise I think it should go to the `prepare` script ? You tell me.

Adel


Tk-Glitch:
Those fs additions are to bypass modprobed-db getting rid of them - at least on Arch -, leading to inaccessible devices for some users. It's not being taken care of by the diet config and will happen regardless of the entries being there, and that's why it was done that way. Not sure if it affects other distros but it's likely.

AdelKS:
That's very weird, so people using modprobeddb or diet should face the same issue. I do not but I'm not on Arch 🤔

AdelKS:
I added the `kernelconfig.new` file to the artifacts so we can check what's really going on, and I will disable the use of the frag file to see. If the frag is still needed, I will extract it to a file and use it for every distro.

I updated the release code so it makes a release as long as one distro successfully builds.

AdelKS:
Okay, the build has finished, and that frag file done for arch doesn't do anything: I removed it and the kernel config file is good, see here https://github.com/AdelKS/linux-tkg/releases/tag/v6.6.10

whitehara:
> Another issue I discovered, is that RPMs are not stripped (that's why they are gigantic), I have no idea how to fix that for now, maybe @whitehara can help me here.

Maybe one of the solution is "Use Fedora official kernel.spec's way".
First, see this sample [kernel.spec](https://download.copr.fedorainfracloud.org/results/whitehara/kernel-tkg/fedora-39-x86_64/06867903-kernel/kernel.spec).
This kernel.spec is used for one of my kernel packages and based on the recent Fedora official kernel.spec.
In this file, you will see these sections.
```
cp_vmlinux()
{
  eu-strip --remove-comment -o "$2" "$1"
}
```

This section strips vmlinux. Before you use this, you need to extract it from vmlinuz.
This may reduce 3-4MB.

```
    find $RPM_BUILD_ROOT/lib/modules/$KernelVer -name "*.ko" -type f >modnames

    # mark modules executable so that strip-to-file can strip them
    xargs --no-run-if-empty chmod u+x < modnames
```

This section will mark module files will be stripped. Because there are a lot of module files in rpm, this may reduce a lot of size.

You can try to integrate these codes into `scripts/package/kernel.spec`.

AdelKS:
@whitehara thank you for your feedback ! Would you be kind to PR the change you are suggesting by changing our https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-patches/6.7/0013-fedora-rpm.patch file, and older versions too ? Like this we merge this MR as-is then we fix fedora.

AdelKS:
@Tk-Glitch let's merge it like this, we can solve Fedora in another PR :D

AdelKS:
Haha I just matrixed the entire thing to build every combination of scheduler and distro, it looks like it's gonna work

Tk-Glitch:
We had the issue before with CI regarding those fs related options, we'll see how it goes. Thanks Adel! Let's merge.

PR #874, opened 2024-01-19
Add PRJC 6.7-r2

sirlucjan:
Update PRJC to latest version


Tk-Glitch:
Thanks sirlucjan :frog: :heart:

PR #879, opened 2024-01-23
Fix winesync patch application for 6.7

krathalan:
This PR is a simple patch that fixes patch application for the winesync patch for kernel 6.7. There was an error in applying the patch to `tools/testing/selftests/Makefile`.


Tk-Glitch:
Thanks krathalan :frog: :heart:

PR #898, opened 2024-03-10
Update PREEMPT_RT's supported kernel subversion and patch revision

blackteahamburger:


Tk-Glitch:
For 5.4 we might not want to update if the patch is still there as we're targeting an older revision than latest (due to some logic conflicts).
Not like it's a real issue though.

Thank you blackteahamburger :frog: :heart:

PR #918, opened 2024-04-08
Several improvements: quick fix

GreyAsteroid:
Noticed install.sh broke due to this being missed.


Tk-Glitch:
Thanks @GreyAsteroid ! :frog: :heart:

PR #926, opened 2024-05-02
6.8 & 6.9: update diet db

AdelKS:
I just compiled a full 6.8 kernel to get bluetooth working (using my own `modprobed-db` db) and just added what was missing our current `minimal-config` files in 6.8 and 6.9


Tk-Glitch:
We'll also want to add the minimum entries for vm usage down the line. Got pm'ed about it not booting on VMs.

Thanks as always Adel! :frog: :heart:

PR #931, opened 2024-05-06
Add Project-C v6.9-r0

whrvt:
Added commits ab3f0a90..ef162b48 from [alfredchen/linux-prjc/linux-6.9.y-prjc](https://gitlab.com/alfredchen/linux-prjc/-/commits/linux-6.9.y-prjc/?ref_type=heads) on top of the linux-tkg [0009-prjc_v6.8-r5.patch](https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-patches/6.8/0009-prjc_v6.8-r5.patch)

This doesn't address _rr_interval, which has been non-functional as a customizable option in linux-tkg for the past few PrjC versions.

Update: tested working on kernel 6.9-rc7 (and -rc6)


Tk-Glitch:
Thanks whrvt! :frog: :heart:

PR #932, opened 2024-05-06
Update customization.cfg

Galcian79:
See https://github.com/sched-ext/scx/issues/269


Tk-Glitch:
Right! Thanks Galcian79! :frog: :heart:

PR #940, opened 2024-05-11
Non-arch: fix logging

AdelKS:
shell-output.log wasn't getting created at all, I was wondering why all the newest bug reports were not attaching it at all and started losing hope in humanity haha


Tk-Glitch:
Nice Adel! Thanks! :frog: :heart: :frog:

Issue #939, opened 2024-05-11
Error 2 on binrpm_pkg when building latest (no rc) on OpenSUSE Tumbleweed


honeymoan:
### System Info

```
VERSION_ID="20240509"
PRETTY_NAME="openSUSE Tumbleweed"
```

Hey,

I'm trying to build tkg on tumbleweed latest but it fails on packaging with a Error 1 then a 2 when running binrpm_pkg.

### Logs
[customization.cfg.txt](https://github.com/Frogging-Family/linux-tkg/files/15284077/customization.cfg.txt)
[prepare.log.txt](https://github.com/Frogging-Family/linux-tkg/files/15284078/prepare.log.txt)
[run.txt](https://github.com/Frogging-Family/linux-tkg/files/15284079/run.txt)
[system-info.txt](https://github.com/Frogging-Family/linux-tkg/files/15284080/system-info.txt)

AdelKS:
Hey can you try again with latest `linux-tkg` commit so you can have one more log file `shell-output.log.txt` ? Thanks

not-NEKO:
maybe i got the same issue don't know much

Issue #941, opened 2024-05-13
OpenSUSE File not found compiling problem


not-NEKO:
Basically after a long time compiling i get this error on 2 differents computer (which run both on openSUSE).
OS: OpenSUSE
What does it do: Doesn't compile
```
Processing files: kernel-6.9.0_tkg_eevdf-1.x86_64
error: File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/kernel
error: File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.builtin
error: File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.builtin.modinfo
error: File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.order

RPM build errors:
    File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/kernel
    File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.builtin
    File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.builtin.modinfo
    File not found: /home/user/Gits/linux-tkg/linux-src-git/rpmbuild/BUILDROOT/kernel-6.9.0_tkg_eevdf-1.x86_64/lib/modules/6.9.0_tkg_eevdf/modules.order
make[2]: *** [scripts/Makefile.package:71: binrpm-pkg] Error 1
[...]

PR #943, opened 2024-05-14
6.9: Update patch for suse

Arias800:
I was able to build the kernel with these changes.
Should fix https://github.com/Frogging-Family/linux-tkg/issues/939 and https://github.com/Frogging-Family/linux-tkg/issues/941 I guess.


Tk-Glitch:
That's great. Thank you very much for Suse users, Arias800.
:frog: :heart:

not-NEKO:
Thank you my bro

PR #954, opened 2024-06-04
remove `grubby` from required Fedora packages

### LINKED ISSUE #952: Install script should not try to install `grubby` on Fedora
**Describe the bug**
 - Which OS do you encounter the bug on?

Fedora 40

 - A clear and concise description of what the bug is.

The install script tries to install the `grubby` package on Fedora. I am not using grub2, but systemd-boot. I cannot install grubby as it conflicts with sdubby (essentially the same thing, but for systemd-boot). Systemd-boot is an official option since Fedora 39.

Anyone using grub2 should have the `grubby` package installed anyway, just like anyone using systemd-boot should have the `sdubby` package. There is no need to try and install it during the build process.

alterNERDtive:
The `grubby` package conflicts with systemd-boot and the corresponding `sdubby` package. [Systemd-boot is an officially supported boot loader since Fedora 39](https://fedoraproject.org/wiki/Changes/cleanup_systemd_install).

Any installation with grub2 has grubby installed by default anyway.

fixes #952

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -81,7 +81,7 @@ _install_dependencies() {
     sudo apt install bc bison build-essential ccache cpio fakeroot flex git kmod libelf-dev libncurses5-dev libssl-dev lz4 qtbase5-dev rsync schedtool wget zstd debhelper ${clang_deps} -y
   elif [ "$_distro" = "Fedora" ]; then
     msg2 "Installing dependencies"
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thanks alterNERDtive! :frog: :heart:

PR #964, opened 2024-07-02
Fix Suse build

Arias800:
Tested on Tumbleweed with Linux 6.10 rc 6.


Tk-Glitch:
Thanks Arias800 ! :frog: :heart:

PR #974, opened 2024-07-16
6.10 pds/bmq

polq1200:
Generated 0009-prjc_v6.10-r0.patch from https://gitlab.com/alfredchen/linux-prjc added pds and bmq options for kernel 6.10, changed 0009-glitched-ondemand-bmq.patch because MICRO_FREQUENCY_MIN_SAMPLE_RATE doesn't exist in 6.10 and the patch was failing. Now it's working.

PS: This is my first contribution to open source. I love the Tkg kernel.


Tk-Glitch:
Thanks polq1200 ! :frog: :heart:

PR #997, opened 2024-09-08
Fix error when applying misc-additions on 6.10.9

polq1200:


Tk-Glitch:
Thanks polq1200! :frog: :heart:

PR #1020, opened 2024-10-05
Add min_base_slice_ns for BORE scheduler

jnines:
CachyOS sets this to 100000, default is 200000
Description reads as:
```
+config MIN_BASE_SLICE_NS
+	int "Default value for min_base_slice_ns"
+	default 2000000
+	help
+	 The BORE Scheduler automatically calculates the optimal base
+	 slice for the configured HZ using the following equation:
+	 
+	 base_slice_ns =
+	 	1000000000/HZ * DIV_ROUNDUP(min_base_slice_ns, 1000000000/HZ)
+	 
+	 This option sets the default lower bound limit of the base slice
+	 to prevent the loss of task throughput due to overscheduling.
+	 
+	 Setting this value too high can cause the system to boot with
+	 an unnecessarily large base slice, resulting in high scheduling
+	 latency and poor system responsiveness.
+
```


ptr1337:
Please revert this again or set it to the default value back. We have changed it, since it had some issues in the past with audio crackling, but this has been fixed now.

Tk-Glitch:
Done

PR #1021, opened 2024-10-09
6.11: sync arch config

kylon:


Tk-Glitch:
Thanks kylon! :frog: :heart:

PR #1024, opened 2024-10-23
tkg-patches: 6.12: Fix glitched-base

hellsgod:
6.12 had a recent change in elevator.c and your gliched-base patch failed to apply because of that.


Nanotwerp:
Fixed by 9ba7f28.

PR #1045, opened 2024-12-09
sync 6.12 arch config

kylon:


AdelKS:
Will be taken care of by #1075

PR #1049, opened 2024-12-14
install.sh: prompt for distro even when using 'config'

### LINKED ISSUE #1025: Genpatches is not applied when configuring
https://github.com/Frogging-Family/linux-tkg/blob/8016959b101deb903cff8a3f3964f530882dc7d1/install.sh#L109-L111
https://github.com/Frogging-Family/linux-tkg/blob/8016959b101deb903cff8a3f3964f530882dc7d1/install.sh#L133-L135

These lines cause genpatches not applied even with `_distro="Gentoo"` when running `./install.sh config`.

AdelKS:
Hopefully I didn't break anything, although I don't think so, tried quickly on my machine.

Fixes #1025

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -102,12 +102,9 @@ fi
 
 if [ "$1" = "install" ] || [ "$1" = "config" ]; then
 
-  if [ -z "$_distro" ] && [ "$1" = "install" ]; then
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Sorry for the delay. I was thinking about doing exactly that. Thanks Adel.
:frog: :frog: :frog: :heart: :heart: :heart:

PR #1053, opened 2024-12-23
Few minor updates

AdelKS:
- Removed the "experimental" flag from "diet" config, works fairly well and I believe we didn't get much issues on that front.
- Updated the CI code (I started working on something similar for wine-tkg, I'll pick it back up soon :tm:)
- Update diet db with few modules
  - Not sure if they apply to earlier or later kernel versions, did not test.

### DIFF
--- a/.github/workflows/build-release-kernels.yml
+++ b/.github/workflows/build-release-kernels.yml
@@ -6,17 +6,17 @@ on:
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


AdelKS:
Just realized the actual kernels are missing from the releases page, need to take a look at it first.

AdelKS:
I will split this PR in separate things

PR #1055, opened 2024-12-26
remove waydroid option, now enabled in arch config

kylon:
see #1045

### DIFF
--- a/.github/workflows/arch-builder.yml
+++ b/.github/workflows/arch-builder.yml
@@ -8,7 +8,6 @@ on:
 
 
 env:
-  _waydroid: "true"
   _processor_opt: "generic"
   _git_mirror: "github.com"
   PKGDEST: "/tmp/linux-tkg"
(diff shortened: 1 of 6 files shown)


AdelKS:
Hey, thanks!

By having a look at arch's new `defconfig`, it has exactly what the script sets. Which is nice. No need to remove the script nor the customization option so the support remains for older kernels and for people who use their own initial `defconfig`.

This PR can therefore be closed, we can maybe put `waydroid=true` by default in the customization as a change

kylon:
i think i ll make it a noop for 6.12+ later today, in case upstream config changes something we don't want to break it

Tk-Glitch:
It's also useful to keep it around for other distros where you can't/don't want to use Arch's defconfig. Turning it to true by default or even always enabling those entries (and get rid of the option) to remove config burden for the user sounds fine.

Tk-Glitch:
So I adopted the last option mentioned above to keep other distros in check, but we might need to fix PDS/BMQ against the current Arch defconfig also. That's a different issue though.
Thank you kylon! :frog: :heart:

kylon:
ok, been busy with other things, took more time than expected

PR #1075, opened 2025-02-10
ci: add workflow to update kernel config

AdelKS:
Little workflow to automatically update kernel configs, one thing less to bother with.


Tk-Glitch:
True true, we can indeed automate this considering we don't rely on editing defconfig by hand anymore.
Thanks Adel :frog: :heart: :heart: :heart:

PR #1081, opened 2025-02-24
Flexible git remotes

AdelKS:
This MR does two things
1. Make `_git_mirror` in `customization.cfg` accept directly git links
2. Make `_version` in `customization.cfg` accept any git tag that is on the remote currently configured.
Needs testing

Fixes #1057

PR #1086, opened 2025-04-09
ci fixes

AdelKS:
It turns out that building with env vars in Arch was somehow always broken, or something changed in `makepkg` recently about env vars, our the PR #1081 subtly changed something, but we've been having issues [like this](https://github.com/Frogging-Family/linux-tkg/actions/runs/14357352030) since.

I took time to look into it, and it was related to the fact that the PKGBUILD gets sourced for each phase, but all the following phases other than the first lose the environment and some important vars get lost. I believe this problem isn't new and that's why the `BIG_UGLY_FROGMINER` file was being used.

I changed the PKGBUILD to dump the env and the config files, in the right order, but only once when the file `BIG_UGLY_FROGMINER` doesn't exist, then the env is always sourced from there and never from elsewhere, the `prepare` script appends things to it litte by little as phases are run. It seems to work nicely.

### DIFF
--- a/.github/workflows/arch-builder.yml
+++ b/.github/workflows/arch-builder.yml
@@ -60,7 +60,7 @@ jobs:
       #     echo "${_modprobeddb_db_path}"
 
       - name: Compile Kernel
-        run: su user -c "yes '' | makepkg --noconfirm -s"
+        run: su --preserve-environment user -c "yes '' | makepkg --noconfirm -s"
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Yeah that's the main (and sometimes sole) reason we are exporting variables to files in all our configurable projects. Nice you catched that one :) Thank you Adel!

PR #1105, opened 2025-05-06
customization: support forwarding user provided compiler flags

AdelKS:
A very good gun to shoot yourself in the foot.

We can also hard code it in the `Makefile`, so even `dkms` modules pick them up. I think DKMS doesn't support passing compilers flags, at least I couldn't find how.

Do we want this in ? I think it works in non-Arch already because the environment doesn't get cleared.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -138,6 +138,9 @@ build() {
       $_ionice -p "$_pid" ||:
     fi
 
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
It's optional, there's a warning, I'm fine with that.

PR #1107, opened 2025-05-17
"patchwork": tool that helps with rebasing patches 

AdelKS:
Added the `patchwork` tool, the arguments it can take
- `print`
  - `unclean`: prints list of patches that don't apply against latest kernel .z version,
                for each x.y kernel version (full version string being x.y.z)
- `rebase`
  - `start [<X.Y/patch-path>|include-eol]` : sets a workfolder with a kernel where you can do any number of extra commits or commit edits on top
                                  Note: the first commit should be left intact so re-creating the patch (after edits) still works.
    - `[no argument]`: sets a workfolder of the first kernel version x.y.z where a patch
                      doesn't apply and starts a 'git am' with that patch.
    - `<x.y/patch-name>` (optional): sets a workfolder of kernel version x.y.z (where .z is the latest) and applies the given patch in argument (after applying the patch it depends on, if there is one)
  - `end` : Rebase done, re-creates (with `git format-patch`) the patch that has been pulled with the user added changes. (without taking into account the parent patch, if there was one)

Added optional `interdeps` folder within each patch folder
- Can contain filenames that have the same name as the patch
  - If it does, the content of that file is a list of patch filenames that it depends on
    - If there's more than one line, it's considered as `xor` (exactly one of them is needed)
- They are applied first before applying the target patch

Then I used it to rebase the various outdated patches, not done yet.
- `prjc`: I rebased directly upstream patches, it's easier/safer to rebase the whole patchset than the squashed version
  -  I stopped syncing those patches from Cachy because of that, and also because they do not maintain the patches for older kernel versions

@Tk-Glitch You must be dealing with even more complex dependency trees in wine, if you think this approach can be transferred to wine & make your life easier, I can look into it.


sebastian-de:
@AdelKS I'm not involved with this project, but I've worked with the OpenWrt build system in the past.
They used `quilt` for patch management - the OpenWrt docs for it are here: https://openwrt.org/docs/guide-developer/toolchain/use-patches-with-buildsystem
Maybe that's something worth looking into and can spare you some work.

Issue #1115, opened 2025-06-02
Update BORE patch for 6.14


kimmot:
### Same-issue check

Yes

### Linux Distribution

Gentoo

### Kernel version

6.14.9

### What happened?

Current BORE patch fails with 6.14.9 kernel:
==> ERROR: An error was encountered applying patches. It was logged to the prepare.log.txt file.

There is new BORE patch available at:
https://github.com/firelzrd/bore-scheduler/blob/main/patches/stable/linux-6.14-bore/0001-linux6.14.9-bore5.9.6.patch

### Relevant logs

Section: bore
Selected: bore
 -> Using bore CPU schduler
 -> Using Arch patches
 -> Applying clear linux patches
 -> Applying glitched base patch
 -> Skipping patch 0003-glitched-base-nonrt.patch...
         (unavailable for this kernel version)
 -> RPM: fixing spec generator
 -> RPM: strip modules
 -> Import Gentoo-kconfig patches
 -> Print firmware names on load
 -> Applying misc additions patch
 -> Skipping patch 0001-mm-Support-soft-dirty-flag-reset-for-VA-range.patch...
         (unavailable for this kernel version)
 -> Skipping patch 0002-mm-Support-soft-dirty-flag-read-with-reset.patch...
         (unavailable for this kernel version)
 -> Applying BORE patch
 **==> ERROR: An error was encountered applying patches. It was logged to the prepare.log.txt file.**
 -> Applying Glitched CFS/EEVDF patch
 -> Using /boot/config-6.14.9-tkg-bore-llvm as config file
 -> Setting cpu micro architecture to znver4
 -> Patching O3 optimization
 -> Patching fsync legacy support
 -> Setting config

### My Favorite type(s) of frog

Frog

mahatmus-tech:
same problem here

sirlucjan:
I've updated BORE: https://github.com/CachyOS/kernel-patches/commit/1a9931bd92c4b557c8c69d6c524772cece3c5843

Could you check it now?

PR #1116, opened 2025-06-02
6.14: Update BORE for tkg kernel users

sirlucjan:
Bugfix for #1115


sirlucjan:
Yes, Ive updated yesterday. Thanks.

PR #1124, opened 2025-06-26
Adding QA code to PKGBUILD.

mabod:
Introducing new variable `_quality_assurance`. If it is set to `true`, PKGBUILD stops at the end of the prepare() function. It will then check `logs/prepare.log.txt` for errors and failures and print 'QA: PASSED' or 'QA: FAILED' accordingly.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -88,8 +88,30 @@ prepare() {
   rm -rf $pkgdir # Nuke the entire pkg folder so it'll get regenerated clean on next build
 
   ln -s "${_kernel_work_folder_abs}" "${srcdir}"
+  
+  if [ "$_quality_assurance" = "true" ]; then
+        export LANG=C
+        myRED='\033[31m'
+        myGREEN='\033[32m'
+        myNC='\033[0m' # No Color
 
-  _tkg_srcprep
+        _tkg_srcprep
+
+        echo
+        echo "####"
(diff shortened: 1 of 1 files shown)


mabod:
The intention of this PR was to prevent the issues with updated but not working patches or with the patchwork directory which all happened recently. The idea was to have an easy way of checking that compilation will not fail upfront as it happened in the past due to bad commits.

Now to your specific questions:

1. This PR resp. the PKGBUILD is agnostic to kernel version. it should work with any kernel version, any scheduler, any linux-tkg config.

2. What exactly do you mean with "makepkg fail earlier". This PR makes  the makepkg process stop right after linux-tkg has run all the scripts and applied all the patches. If it fails that early the assumption is that it is logged in logs/prepare.log.txt.

3. This PR is only for Arch based distros because it is implemented as a PKGBUILD functionality. 

4. It was not my intention to work on a "100% automated approach". I wanted to provide functionality for developer like you that help you to test your commits befor you push them to the repo. In the past we had several commits breaking the build process. The idea is to provide a quick & dirty QA check before a commit is pushed to github.

Anyways, I am not a developer. This is in fact my first PR ever. It is just a proposal. Feel free to canclel the PR and do something completely different.

AdelKS:
Sorry this is not useful at all, pressing CTRL+C when the kernel starts building has the same effect and takes the same time.

PR #1134, opened 2025-07-11
debian/ubuntu: libncurses5-dev deprecation

dacampsss:
linux-tkg wasn't compiling in my updated system Debian 13 trixie. So I just edited libncurses5-dev to libncurses-dev then it compiled again. libncurses5-dev package is deprecated now for both Debian 12 and 13. As for Ubuntu, I tested installing libncurses-dev in my Ubuntu 24.04 VM and it worked good, so I guess it'll work in any recent Ubuntu (LTS) or Debian, and fix the issue for Debian 13 (testing for now, but coming to be stable really soon).


AdelKS:
Thank you very much for your thorough testing !

PR #1153, opened 2025-10-03
Change header level for Linux-tkg section

damachine:
Updated the header from '## linux-tkg' to '# Linux-tkg' for improved formatting.

Hey, great, dear master. The awesome Linux TKG package should be one size bigger ;) A side effect is that my little TKG installer wrapper script looks better in the preview info and also matches the other packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## linux-tkg
+# Linux-tkg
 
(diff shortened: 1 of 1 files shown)


damachine:
> not sure about the capital L in the beginning, do you really need that ? :P

Of course not. Purely cosmetic! :D

damachine:
but this could increase performance :O

AdelKS:
> but this could increase performance :O

Well... if we're talking performance improvements, then yeah it should be added !

I thought all other tkg projects don't start with a capital letter in their readme, but it seems to be a mix and match, I guess we can even go with : "Linux-tkg: build your own kernel" to match more with the other projects

PR #1154, opened 2025-10-04
make _debugdisable more aggressive

AdelKS:
Need to try a kernel built with this first.

Input welcome, @sirlucjan  maybe you have some thoughts on the matter


Tk-Glitch:
Looks fine to me. We were quite conservative. Merging!

Issue #1157, opened 2025-10-22
Set defaults for _debugdisable_disabled_entry and _debugdisable_enabled_entry


Ruttie2006:
### Same-request check

Yes

### Linux Distribution affected

All

### Feature description

Currently if `_debugdisable_disabled_entry` and `_debugdisable_enabled_entry` are not set in customization.cfg, nothing is stripped due to `_debugdisable_disabled_entry` not being set to a reasonable default.
Relevant line(s): https://github.com/Frogging-Family/linux-tkg/blob/e768f4d1ac76da18303e3a1c5c9c78a16873801b/linux-tkg-config/prepare#L1111
It would be nice if some default was set if the option(s) are not set, this would allow for a cleaner customization.cfg for those who simply want to accept default options.

### My Favorite type(s) of frog

African Dwarf Frog (Hymenochirus)

Issue #1156, opened 2025-10-17
SCX can't start schedulers. Permission denied and futex error


WhitePeace36:
### Same-issue check

Yes

### Linux Distribution

Archlinux

### Kernel version

6.17.2 and 6.17.3

### What happened?

When trying to start a scheduler with scx_loader or another method, the startup failed. 

There are 2 errors the 1. is some futex bpf error and the 2. is a permission error.

For the permission error i found out that it has something todo with bpf_rcu_read_lock and the other calls from bpf whch are not available for some reason.

when i type : 
sudo bpftool feature | grep rcu -i
i don't get any output. so I assume there is nothing to call.

maybe there is a new option in the kernel config to activate this calls.

### Relevant logs

[scx_logs_futex.log](https://github.com/user-attachments/files/22978172/scx_logs_futex.log)

[scx_permission_error.log](https://github.com/user-attachments/files/22978258/scx_permission_error.log)

### My Favorite type(s) of frog

Frog

WhitePeace36:
i tested it out. removing FTRACE from the ${_debugdisable} option in /linux-tkg/linux-tkg-config/prepare does the trick it seems.

But i would remove FTRACE, SCHED_INFO, SCHEDSTATS and RCU_TRACE from the ${_debugdisable} . Just to be safe.

AdelKS:
Hey, 

Thanks for tracking down the root cause of your issue, I'll put back those CONFIG options.

Weirdly enough, we keep all [these options](https://github.com/sched-ext/scx/blob/main/kernel.config) untouched, but oh well, I guess it's not up-to-date.

AdelKS:
By looking more closely at the options linked above (that's what the official docs say we should have), just restoring `FTRACE` will give back everything in that file. Since `_debug_disable` is basically saying "I want the least amount of overhead due to kernel debugging", the we can still ignore `SCHEDSTATS` and `SCHED_INFO`. `RCU_TRACE` seems to be entirely unrelated.

```
config SCHEDSTATS
	bool "Collect scheduler statistics"
	depends on PROC_FS
	select SCHED_INFO
	help
	  If you say Y here, additional code will be inserted into the
	  scheduler and related routines to collect statistics about
	  scheduler behavior and provide them in /proc/schedstat.  These
	  stats may be useful for both tuning and debugging the scheduler
	  If you aren't debugging the scheduler or trying to tune a specific
	  application, you can say N to avoid the very slight overhead
	  this adds.
[...]

PR #1169, opened 2025-12-06
customization: make _ftracedisable a nuclear option

AdelKS:
Entirely disables FTRACE and all its children, I am assuming those of us who disable FUNCTION_TRACER & GRAPH_TRACER actually don't\ need FTRACE altogether.

For the sched-ext issue #1156:
- it seems that FTRACE is needed by at least LAVD
  - issue open here https://github.com/sched-ext/scx/issues/2952
- Got fixed upstream for LAVD to still work with FTRACE disabled but through a less optimal path

What do you think about that @Tk-Glitch ?

Closes: #1157
Closes: #1156


Tk-Glitch:
Nuclear sounds okay in this specific context. The note is clear. I'm fine with the change.

PR #1180, opened 2026-02-09
Update to enable clang cross compile

Bacon-O:
clang-cross-compile

In customation added _llvm_target_arch=""

Configure prepare to read _llvm_target_arch

blank --> no change/default

"x86_64-linux-gnu" --> enables cross compiling for x86_64-linux-gnu

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -120,6 +120,10 @@ _cpusched=""
 # For advanced users.
 _compiler=""
 
+# [Advanced] LLVM target architecture (e.g. "x86_64-linux-gnu").
+# Leave empty to use default.
+_llvm_target_arch=""
+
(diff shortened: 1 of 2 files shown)


Tk-Glitch (approved):

Bacon-O:
> Hey!
> 
> So, from what I am understanding, the `-march=foo` check if `foo` applies to _your_ machine makes cross-compiling not work since, well, it's not _your_ machine. And you are adding a workaround for that.
> 
> What else do you need to edit for cross compiling to work with `linux-tkg` ?

I was looking at this a while ago and yes there was more I had to set. I was cross compiling X86 on ARM64. It was a finicky environment I had to setup.

This is on my todo list to revisit so when I do will update what else is needed.

PR #1192, opened 2026-04-04
pkgbuild: update makedepends and optdepends

damachine:
hey all, thanks for the backstory @ TkG, good to know the reasoning behind it all! figured i'd just go ahead and open the pr for the points `makedeps` ad `optdeps` since those seemed fine. left the `provides` thing alone for now since it sounds like there's history there :P  having a lot of fun contributing here, hope i'm not being too pushy at all haha

I'm not entirely sure about the other makedeps; I included them from the arch pkgbuild for the sake of completeness.
ref: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/PKGBUILD?#L12

a little llm comic – just for the fun of it, for all you crazy little nerd frogs 🐸🐧 meant with love :)
pr_linux_tkg

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -67,9 +67,12 @@ pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="https://www.kernel.org/"
 license=('GPL2')
-makedepends=('bison' 'xmlto' 'docbook-xsl' 'inetutils' 'bc' 'libelf' 'pahole' 'patchutils' 'flex' 'python-sphinx' 'python-sphinx_rtd_theme' 'graphviz' 'imagemagick' 'git' 'cpio' 'perl' 'tar' 'xz' 'wget')
+makedepends=('bc' 'bison' 'cpio' 'docbook-xsl' 'flex' 'git' 'graphviz' 'imagemagick' 'inetutils' 'libelf' 'pahole' 'patchutils' 'perl' 'python-sphinx' 'python-sphinx_rtd_theme' 'tar' 'wget' 'xmlto' 'xz')
+if [[ $_kver -gt 600 ]]; then  
+  makedepends+=('gettext' 'python' 'rust' 'rust-bindgen' 'rust-src' 'zstd')
+fi
 if [ "$_compiler_name" = "-llvm" ]; then
-  makedepends+=( 'lld' 'clang' 'llvm')
+  makedepends+=('clang' 'llvm' 'lld')
 fi
 optdepends=('schedtool')
 options=('!strip' 'docs')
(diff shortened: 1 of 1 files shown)


PKGBUILD
```diff
@@ -67,9 +67,16 @@ pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="https://www.kernel.org/"
 license=('GPL2')
-makedepends=('bison' 'xmlto' 'docbook-xsl' 'inetutils' 'bc' 'libelf' 'pahole' 'patchutils' 'flex' 'python-sphinx' 'python-sphinx_rtd_theme' 'graphviz' 'imagemagick' 'git' 'cpio' 'perl' 'tar' 'xz' 'wget')
+makedepends=('bc' 'bison' 'cpio' 'docbook-xsl' 'flex' 'git' 'graphviz' 'imagemagick' 'inetutils' 'libelf' 'pahole' 'patchutils' 'perl' 'python-sphinx' 'python-sphinx_rtd_theme' 'tar' 'wget' 'xmlto' 'xz')
+case $_basever in
```
AdelKS: Better go with something like
```
if [[ $_basever -gt 600 ]]; then
```

basically kernel `X.YY` is mapped to `XYY`, and `X.Y` -> `X.0Y` (like `6.1` is `601`)

damachine:
> Crazy how AI can make that in just a few seconds...

Yeah, that's crazy, but also kind of scary.

PKGBUILD
damachine: Yeah that looks much smoother. Should we apply the same pattern to the case block in hackheaders() as well? like
```
if [[ $_basever -gt 510 ]]; then
  depends=('pahole')
fi
```

AdelKS: Yeah go ahead!

And sorry, it's actually [_kver](https://github.com/Frogging-Family/linux-tkg/blob/master/linux-tkg-config/prepare#L247) (I always confuse which is which)

PKGBUILD
```diff
@@ -67,9 +67,12 @@ pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="https://www.kernel.org/"
 license=('GPL2')
-makedepends=('bison' 'xmlto' 'docbook-xsl' 'inetutils' 'bc' 'libelf' 'pahole' 'patchutils' 'flex' 'python-sphinx' 'python-sphinx_rtd_theme' 'graphviz' 'imagemagick' 'git' 'cpio' 'perl' 'tar' 'xz' 'wget')
+makedepends=('bc' 'bison' 'cpio' 'docbook-xsl' 'flex' 'git' 'graphviz' 'imagemagick' 'inetutils' 'libelf' 'pahole' 'patchutils' 'perl' 'python-sphinx' 'python-sphinx_rtd_theme' 'tar' 'wget' 'xmlto' 'xz')
+if [[ $_kver -gt 600 ]]; then
```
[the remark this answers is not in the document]
damachine: Yes, If I saw that correctly _kver for 6.0 = 600, 6.1 = 601, so -gt 600 triggers exactly from 6.1.

Rust support was integrated into 6.1 
https://www.phoronix.com/news/Rust-Is-Merged-Linux-6.1

Issue #1193, opened 2026-04-11
Option to compile without Rust


PootisSpencer:
### Same-request check

Yes

### Linux Distribution affected

All

### Feature description

How about adding an option for people who don't want Rust on their systems?

It would be a boolean that would live in customization.cfg

### My Favorite type(s) of frog

Frog, African Bullfrog (Pyxicephalus adspersus)

AdelKS:
You don't want `linux-tkg` to pull `rust` dependencies ? 

Currently it's only done on Arch, and afaik we actually don't build rust components at all.

damachine:
Btw. The `rust` makedepend for Arch doesn't change anything about the actual kernel build.

If you really don't want rust at all, you'd need to set (CONFIG_RUST) in your config.

PootisSpencer:
If in the case of Arch it's pulled as makedepends for no particular reason, then this request may be couple years too early.

PR #1194, opened 2026-04-11
pkgbuild: update "makedepends" with upstream

damachine:
Hi all. 

Cleaned up some makedepends (zstd, gettext) carried over from the last PR. picked up from arch pkgbuild but that aren't needed for rust. Maybe we should look again at why we didn't have them in there.

KISS

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -67,10 +67,27 @@ pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="https://www.kernel.org/"
 license=('GPL2')
-makedepends=('bc' 'bison' 'cpio' 'docbook-xsl' 'flex' 'git' 'graphviz' 'imagemagick' 'inetutils' 'libelf' 'pahole' 'patchutils' 'perl' 'python-sphinx' 'python-sphinx_rtd_theme' 'tar' 'wget' 'xmlto' 'xz')
-if [[ $_kver -gt 600 ]]; then  
-  makedepends+=('gettext' 'python' 'rust' 'rust-bindgen' 'rust-src' 'zstd')
-fi
+makedepends=(
+  bc
+  binutils
+  cpio
+  gettext
+  glibc
+  libelf
+  libgcc
+  openssl
+  pahole
+  perl
+  python
+  rust
+  rust-bindgen
+  rust-src
+  tar
+  xxhash
+  xz
+  zlib
+  zstd
+)
 if [ "$_compiler_name" = "-llvm" ]; then
   makedepends+=('clang' 'llvm' 'lld')
 fi


damachine:
Okay, as discussed @AdelKS, removing the extra .cfg for Rust.  

What do you think. should we remove the other makedepends (gettext, zstd). They have nothing to do with Rust, and I think we only want to handle Rust here in `-gt 600`. As mentioned above, I had them in my last PR mistakenly. Otherwise i think, we can close this :P

AdelKS:
Thanks! 

1. Can you edit your first message to remove the `Fix #1193` so it doesn't close that issue ? Now this PR is just about fixing something that got mistakenly introduced, if I understood correctly :D
2. Can you drop the `merge-branch` commit ?

AdelKS:
Oh yeah, the commit message isn't accurate anymore, needs a litte `git commit --amend` or `git-reword` :P

damachine:
> Now this PR is just about fixing something that got mistakenly introduced, if I understood correctly :D

mea culpa xD

AdelKS:
Now that I looked at the upstream Arch [latest PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/PKGBUILD), and compared it to the [one from 5.19](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/5.19.13.arch1-1/PKGBUILD), we're just basically outdated.

I think we should overwrite with the version with latest PKGBUILD, minus the docs
```
makedepends=(
  bc
  binutils
  cpio
  gettext
  glibc
  libelf
  libgcc
  openssl
  pahole
  perl
  python
  rust
  rust-bindgen
  rust-src
  tar
  xxhash
  xz
  zlib
  zstd
)
```

Actually, no need for a condition for checking before or after linux 6.0, because the dependencies should refer to latest state of the Archlinux packages, older kernels will remain buildable I think (otherwise it is a pain to track anyway)

AdelKS:
> mea culpa xD

No no, you were on the right path, turns out our dependencies are simply carried over from the past hehe

damachine:
> Maybe we should look again at why we didn't have them in there.

Yeah, my words :D

damachine:
We let rust in -gt, right?

AdelKS:
> We let rust in -gt, right?

I don't think it's useful in the end: it prevents from pulling rust when building old kernels ? Let's just follow latest deps from upstream and not add unnecessary complexity in the end. 

I didn't look at this deeply enough last time sorry.

damachine:
> I don't think it's useful in the end: it prevents from pulling rust when building old kernels ? Let's just follow latest deps from upstream and not add unnecessary complexity in the end.

I see it exactly the same way. If the rust config doesn't exist in the kernel config, there shouldn't be any problems. the makedeps will not change the kernel itself.

~okay, is this `options` relevant for docs?~

damachine:
> okay, is this `options` relevant for docs?

It looks like this isn't relevant. 

~I don't know if we even need `docs` in `options`. Maybe we can save some space, haha.~

Okay, sorry, I see I'm wrong, that has nothing to do with the Kenel docs. we need this for 

`/usr/share/doc/${pkgbase}/customization.cfg"`

That's enough of that tangent, sorry.

Back to business:
As mentioned above, the new deps (rust, etc.) shouldn't cause any problems as they don't change anything in the kernel build. I can't say if we need "git" or something like that for linux-tkg build in some point. ;)

Otherwise I think we can do it like this way

AdelKS:
Thanks for bringing this up ! You are actively helping removing some crust from `linux-tkg` hehe :frog:

PR #1195, opened 2026-04-15
patch: 6.0-7.0: refresh 0013-optimize_harder_O3

damachine:
Regenerate patches against latest prepare KCONFIG logic for O3.

- 6.18 6.19 7.0 update with cachyos upstream changes https://github.com/CachyOS/linux/commit/23af7316ce8dec93b00bbf22b7343f499c04fae9


damachine:
To safty finish. then we can close this O3 thing ;)

I think we should take the same new O3 patches for all from 6.0 so that the KCONFIG is activated correctly in the prepare file. The old patches still have the old logic.

it won't affect many users, but for the sake of completeness. 

KISS

AdelKS:
Please try the `patchwork` script that's in linux-tkg-patches folder to check that at least the patches cleanly apply. You will find this tool useful for any future contribution / maintenance of patches hehe

damachine:
> Please try the `patchwork` script that's in linux-tkg-patches folder to check that at least the patches cleanly apply. You will find this tool useful for any future contribution / maintenance of patches hehe

Thanks!
Haha, I didn't know that; I built something similar myself xD

Ah, just to clarify, the patches basically change more than just -O3; you could say they're "compiler optimizations" Nevertheless, I think we can leave it as is(Name). Of course, there's no rust in 6.0, which is different.

Okay `patchwork` test was finish with a beautiful `All patches apply cleanly!`

AdelKS:
Thanks! :frog:

PR #1201, opened 2026-04-26
pkgbuild: add KSMBD to provides array, sync header depends

damachine:
Hi! Some parts of the PKGBUILD still seemed a bit stuck in the "old way", so I wanted to freshen things up a bit. I hope these changes feel right!

PKGBUILD: Update provides and depends

- **hackbase()**: Added `ksmbd` to the `provides` array. 
This module has been integrated into the mainline kernel since version 5.15 and should be listed for userspace tools.
- **hackheaders()**: Synced depends with current Arch/CachyOS standards.
Removed the `-gt` version comparison. Is there still a reason to keep this?

Refs:
* [KSMBD Mainline Mergelog (v5.15)](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ksmbd?h=v5.15)
* [Upstream PKGBUILD Reference](https://gitlab.archlinux.org/archlinux/packaging/packages/linux-zen/-/blob/main/PKGBUILD?ref_type=heads#L119)

<3

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -183,9 +183,9 @@ hackbase() {
               'update-grub: Simple wrapper around grub-mkconfig.'
               'scx-scheds: to use sched-ext schedulers')
   if [ -e "${srcdir}/ntsync.rules" ]; then
-    provides=("linux=${pkgver}" "${pkgbase}" VIRTUALBOX-GUEST-MODULES WIREGUARD-MODULE NTSYNC-MODULE ntsync-header)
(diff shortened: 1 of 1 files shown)


AdelKS:
Thanks! 

> Removed the -gt version comparison. Is there still a reason to keep this?

No reason I think. Since we need to follow _current_ Archlinux packaging state even if building older kernels (unless older kernels needs something _more_)

PR #1202, opened 2026-04-28
fix: use portable shebangs for non-FHS compatibility

ashisgreat22:
Changed `#!/bin/bash` to `#!/usr/bin/env bash` in `install.sh`

`#!/bin/bash` assumes bash lives at `/bin/bash`, which is not true on non-FHS compliant distros like NixOS. `#!/usr/bin/env bash` uses `env` to locate bash from `$PATH`, making the scripts portable across all Linux distros.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


AdelKS:
thanks!

PR #1203, opened 2026-04-30
linux-tkg: drop kernel-6.19 to EOL, refresh pkgbuild, minor adjustment

damachine:
Hey, I don't think I need to say much about this.

We seem to run this twice for Arch :thinking:
https://github.com/Frogging-Family/linux-tkg/blob/00dfead1cd6eb12e7a8df05b041160fef810b890/linux-tkg-config/prepare#L2029

This should suffice here:
https://github.com/Frogging-Family/linux-tkg/blob/00dfead1cd6eb12e7a8df05b041160fef810b890/linux-tkg-config/prepare#L1925

I know this is a usermod parameter and rather minor, but I noticed it after adding the linux-tkg logging to the nvidia-all logic. If I haven't completely missed something (spaces) xD, this should be okay.

Other small changes:

- PKGBUILD hackbase(): Remove crda from optdepends (it's no longer in the Arch repo) and added $pkgname-headers.
- install.sh: Only enable ccache when the binary is available (matching PKGBUILD logic).
- prepare: Move 6.19 to EOL list

Take care. Peace <3

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -66,7 +66,7 @@ pkgrel=273
 pkgdesc='Linux-tkg'
 arch=('x86_64') # no i686 in here
 url="https://www.kernel.org/"
-license=('GPL2')
+license=('GPL-2.0-only')
 makedepends=(
   bc
   binutils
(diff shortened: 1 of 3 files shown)


AdelKS:
Thanks ! :frog:

PR #1204, opened 2026-05-01
add '_vanilla' toggle: do not add any* extra patch

AdelKS:
Hey @Tk-Glitch  and all,

I was about to report something upstream about `x2apic`, turns out it's not a linux bug, but I think this toggle can be useful sometimes.

The approach I suggest is to simply affect `_tkgpatcher` to make it refuse patching unless it is told that it's a "vanilla patch". All the CONFIG changes that rely on patches will be written then cleaned up later by `make defconfig` so it shouldn't matter ?

Feedback welcome.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -24,10 +24,12 @@ fi
 
 source "$_where"/BIG_UGLY_FROGMINER
 
+_vanilla_tag=""
+_vanilla_mode && _vanilla_tag="vanilla-"
 if [ -n "$_custom_pkgbase" ]; then
   pkgbase="${_custom_pkgbase}"
 else
-  pkgbase=linux"${_basever}"-tkg-"${_cpusched}"${_compiler_name}
+  pkgbase="linux${_basever}-tkg-${_vanilla_tag}${_cpusched}${_compiler_name}"
 fi
 pkgname=("${pkgbase}" "${pkgbase}-headers")
 pkgver="${_basekernel}"."${_sub}"
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -166,6 +166,10 @@ _module() {
   done
 }
 
+_vanilla_mode() {
+  [[ "$_vanilla" =~ ^(Y|y|Yes|yes|true|1)$ ]];
+}
+
 _prompt_from_array() {
   # Prompts from array, selects default index on empty user input
   # Set the _default_index variable to enable default index selection
@@ -389,8 +393,9 @@ _set_cpu_scheduler() {
     _avail_cpu_scheds+=("muqss")
   fi
 
-  if [ "${_preempt_rt}" = "1" ]; then
-    warning "! Since you have enabled _preempt_rt, incompatible cpu schedulers will not be available !"
+  if [[ "${_preempt_rt}" == "1" ]] || _vanilla_mode; then
+    [[ "${_preempt_rt}" == "1" ]] && warning "! Since you have enabled _preempt_rt, incompatible cpu schedulers will not be available !"
+    _vanilla_mode && warning "_vanilla toggle active, only default scheduler available"
     _avail_cpu_scheds=("$_default_cpu_sched")
   fi
 
@@ -673,7 +678,7 @@ _tkg_initscript() {
 user_patcher() {
     for _f in "$_where"/*."${_userpatch_ext}patch" "$_where"/*."${_userpatch_ext}revert"; do
       if [ -e "${_f}" ]; then
-        warning "Found userpatch file ${f##*/} in the PKGBUILD directory." #"
+        warning "Found userpatch file ${_f##*/} in the PKGBUILD directory." #"
         warning "Userpatches must now be placed in version-specific subdirectories (linux${_basever}-tkg-userpatches for this kernel version)."
         warning "The patch will not be applied."
       fi
@@ -684,6 +689,7 @@ user_patcher() {
 	# To patch the user because all your base are belong to us
 	local _patches=("$_where"/linux"$_basever"-tkg-userpatches/*."${_userpatch_ext}revert")
 	if [ ${#_patches[@]} -ge 2 ] || [ -e "${_patches}" ]; then
+    _vanilla_mode && warning "Vanilla mode is enabled, you may not want to apply some of these user revert patches."
 	  if [ "$_user_patches_no_confirm" != "true" ]; then
 	    msg2 "Found ${#_patches[@]} 'to revert' userpatches for ${_userpatch_target}:"
 	    printf '%s\n' "${_patches[@]}"
@@ -706,6 +712,7 @@ user_patcher() {
 
 	_patches=("$_where"/linux"$_basever"-tkg-userpatches/*."${_userpatch_ext}patch")
 	if [ ${#_patches[@]} -ge 2 ] || [ -e "${_patches}" ]; then
+    _vanilla_mode && warning "Vanilla mode is enabled, you may not want to apply some of these user patches."
 	  if [ "$_user_patches_no_confirm" != "true" ]; then
 	    msg2 "Found ${#_patches[@]} userpatches for ${_userpatch_target}:"
 	    printf '%s\n' "${_patches[@]}"
@@ -728,6 +735,10 @@ user_patcher() {
 }
 
 _tkg_patcher() {
+  if _vanilla_mode && [[ "$1" != "--vanilla-friendly" ]]; then
+    warning "_vanilla toggle active: not applying ${tkgpatch##*/}"
+    return 0
+  fi
   if [ -e "$tkgpatch" ] && [[ $(wc -l <"$tkgpatch") -ge 7 ]]; then
     msg2 "$_msg"
     echo -e "### Applying ${tkgpatch##*/}... ###" >> "$_where"/logs/prepare.log.txt
@@ -748,16 +759,15 @@ _tkg_srcprep() {
   fi
 
   if [ "${_distro}" = "Arch" ]; then
+    local _vanilla_tag=""
+    _vanilla_mode && _vanilla_tag="vanilla-"
     if [ -n "$_custom_pkgbase" ]; then
-      if [[ "$_custom_pkgbase" != *tkg* ]]; then
-        _localversion_tag="-tkg-"
-      else
-        _localversion_tag="-"
-      fi
-      echo "-${pkgrel}${_localversion_tag}${_custom_pkgbase}" > localversion.10-pkgrel
+      local _tkg_watermark=""
+      [[ "$_custom_pkgbase" != *tkg* ]] && _tkg_watermark="tkg-"
+      echo "-${pkgrel}-${_tkg_watermark}${_vanilla_tag}${_custom_pkgbase}" > localversion.10-pkgrel
       echo -e "Version tail set to \"-${pkgrel}${_localversion_tag}${_custom_pkgbase}\"\n" > "$_where"/logs/prepare.log.txt
     else
-      echo "-$pkgrel-tkg-${_cpusched}${_compiler_name}" > localversion.10-pkgrel
+      echo "-$pkgrel-tkg-${_vanilla_tag}${_cpusched}${_compiler_name}" > localversion.10-pkgrel
       echo -e "Version tail set to \"-$pkgrel-tkg-${_cpusched}${_compiler_name}\"\n" > "$_where"/logs/prepare.log.txt
     fi
     echo "" > localversion.20-pkgname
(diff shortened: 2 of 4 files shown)


linux-tkg-config/prepare
```diff
@@ -682,11 +683,15 @@ user_patcher() {
 }
 
 _tkg_patcher() {
+  if [[ "$_vanilla" =~ ^(Y|y|Yes|yes)$ && ! "$_vanilla_patch" =~ ^(Y|y|Yes|yes)$ ]]; then
+    warning "_vanilla toggle active: not applying ${tkgpatch##*/}"
+  fi
```
AdelKS, line 741: hehe turns out I didn't test it correctly as I forgot to put a `return` here...

damachine:
Hey Adel and all, nice idea I actually had the same itch and built a similar thing,  I'm definitely in favor of doing something like this, even if it's not as trivial as it first seems. I've tested it and finished it for mysef. I could needed it for nvidia-all. So let me share what I ran into. 

---

**`_vanilla_patch="y"` vs `--vanilla-ok`**

Your approach of setting a global `_vanilla_patch="y"` before specific `_tkg_patcher` gave me the follwing idee

I went with a `--vanilla-ok` argument pattern instead. the intent lives at the call site:

```bash
_tkg_patcher() {
  local _vanilla_ok=false
  [[ "$1" == "--vanilla-ok" ]] && { _vanilla_ok=true; shift; }
  if _vanilla_active && ! $_vanilla_ok; then
    warning "Vanilla mode active: skipping ${tkgpatch##*/}"
    return 0
  fi
  ...
}

# call site — explicit, no global state
tkgpatch="$srcdir/0013-fedora-rpm.patch"
_msg="RPM: fixing spec generator" && _tkg_patcher --vanilla-ok
```

Same pattern for `_enable`/`_disable`/`_module`/`_tkg_setconfig` — they all guard themselves, opt-in exceptions are declared inline.

---

**`make defconfig` will break your day**

I actually built with `make defconfig` and got this:

```
ERROR: modpost: "v4l2_format_info" [v4l2loopback.ko] undefined!
ERROR: modpost: "video_ioctl2" [v4l2loopback.ko] undefined!
...
```

`defconfig` is extremely minimal, a bunch of common drivers, basically anything not strictly needed to boot gets disabled.

What could works: copy the distro config as usual, then run `make olddefconfig` on top. That keeps the full config (all drivers intact) and just updates any new Kconfig symbols to their upstream defaults. This should works for both Arch and non-Arch if I see this right:

```bash
if [ -z "${_configfile}" ]; then
  # Arch: use distro config as base
  cat "${srcdir}"/config.x86_64 > ./.config
  if _vanilla_active; then
    msg2 "Vanilla mode: running make olddefconfig"
    make olddefconfig ${llvm_opt}
  fi
elif [ "${_configfile}" = "running-kernel" ]; then
  # non-Arch: use running kernel config as base
  cp /boot/config-$(uname -r) .config  # or /proc/config.gz fallback
  if _vanilla_active; then
    msg2 "Vanilla mode: running make olddefconfig"
    make olddefconfig ${llvm_opt}
  fi
fi
```

---

**`_vanilla_active()` helper**

Instead of repeating `[[ "$_vanilla" =~ ^(Y|y|Yes|yes)$ ]]` in 10 places, one function keeps it DRY and makes the intent obvious:

```bash
_vanilla_active() {
  [[ "$_vanilla" =~ ^(Y|y|Yes|yes|true|1)$ ]];
}
```

Also added `true` and `1` — people write configs in all kinds of ways. :P

---

To make things easier for myself, I'm also directly overriding the `.cfg` parameters. It doesn't look very clean, and I'm not even sure if all of that matters in vanilla.

---

Sorry if this is too much or if I'm wrong, and if it looks confusing. xD
If any I can help test I am your wingman ;)

Anyway, the core idea is solid. Happy to compare notes if useful.

AdelKS:
Hey,

Thanks for your feedback !

- `_vanilla_active()` is good stuff, I like it.
- `--vanilla-ok` good! Also leaner, I like it!
- `defconfig` sorry I had in mind `olddefconfig`, we're good with that if we `_enable` a config that doesn't exist ? e.g. a config that's introduced by a patch that's not vanilla
  - I think the AI you used missed that we are already doing what it's suggesting: look at the use of the `_config_updating` variable.

damachine:
Hehe, yeah, busted. Yeah, these things do save us a lot of work sometimes, xD hehe. But of course, you're at an advantage if you know how the linux-tkg works.

AdelKS:
I think LLMs can be super useful ! I use them as a "documentation" when dipping toes in new code bases (so I can ask questions about the code and let it answer) or things I've never done (like Qt + Shaders lately).

I personally don't have issues being at the receiving end of AI assisted contributions as long as it doesn't fully boil down to "I can do it myself with my own LLM, don't need to talk to yours through a PR, otherwise you can donate so I can use more tokens" xD

AdelKS:
Are you happy with my changes ? Everything okay ?

damachine:
What about the package name tag? Should it have a "vanilla" tag? You've already configured it that way in the installsh, or am I missing something?

AdelKS:
> What about the package name tag? Should it have a "vanilla" tag? You've already configured it that way in the install.sh, or am I missing something?

Yeah I did, ah I didn't touch the PKGBUILD

linux-tkg-config/prepare
```diff
@@ -708,6 +717,9 @@ _tkg_srcprep() {
       else
         _localversion_tag="-"
       fi
+      if _vanilla_mode; then
```
damachine: sry if i see it right you have it here. think the pkgbuild is wrong

AdelKS: I think that wasn't enough, I just made a change , can you try an empty modprobeddb build to check ? (so it goes fast)

PKGBUILD
```diff
@@ -55,10 +55,12 @@ fi
 
 source "$_where"/BIG_UGLY_FROGMINER
 
+_vanilla_tag=""
+vanilla_mode && _vanilla_tag="vanilla-"
```
damachine: Ah, sorry, that was a typo, not a source problem. The `_` is missing in `vanilla_mode`; it should be `_vanilla_mode` 😄

AdelKS: Ah! Fixed the same typo elsewhere and missed this one. Works after this fix ?

damachine:
```
Prepared linux-tkg version 7.0.5-273-vanilla-linux-tkg-dirty
==> Beginne build()...
```

build is fine. all LGTM!

Ah, hehe, okay, sry it worked before(because the _localversion_tag in the prepare). 
You wanted to change the release name only (uname -r). I edited the `custom_pkgname`, and I thought that would change (pkgname-vanilla) too, lol. I think it's fine now. We want to change the pkgname only if not empty `-n` :+1: 

and also as an afterword. I completely agree with you about the LLMs! The Qt shader world sounds exciting, btw 😄

AdelKS: So we need to keep the PKGBUILD change too right ?

damachine: > So we need to keep the PKGBUILD change too right ?

I would leave it as is; if custom_pkgname is empty, than `vanila` makes sense in the pkgname.

AdelKS:
I think we are good to go, but will let it marinate a few more days still.

damachine:
It's good that you're giving it more time. I went into it a little deeper. 

To put it short/simply, I ask me what about:

```
_STRIP
K..FLAGS
_community_patches
_user_patches	
```

don't we need a vanilla default(guard) for these?

sry if I ask I'm not so sure how a vanilla kernel should be built.

AdelKS:
Thanks for thinking more about it.

I defined vanilla as "no patch to the kernel sources":
- Currently user patches won't go through, but maybe we should allow them as "vanilla friendly" because they can be patches one would try above vanilla to test stuff :thinking: So a warning is better than not allowing it.
  - Community patches, I think we can keep on not
- `_strip`: no problem
- KFLAGS: no problem, as `make olddefconfig` will remove unknown CONFIG that are potentially introduced by patches but activated anyway with `_enable/_disable/...` (because we didn't touch that in this PR)

damachine:
Thank you for your time and effort!

I was curious to hear your opinion on this again. That's exactly what I was thinking, but I wasn't sure.

So:

> Currently, user patches won't go through, but maybe we should allow them as "vanilla friendly" because they can be patches one would try above vanilla to test things 🤔 So a warning is better than not allowing it. Community patches, I think we can keep on

Completely agree :100:, a warning is the right call there.

Nothing else jumping out at me right now, but yeah let it marinate a few more days.

linux-tkg-config/prepare
damachine: For the sake of elegance, I’ve also inserted this in between:

```bash
if _vanilla_mode; then
  if [[ "$_custom_pkgbase" != *vanilla* ]]; then
    _localversion_tag+="vanilla-"
  fi
fi
```

This ensures that "vanilla" isn't duplicated if you are already tagging your kernel that way. much like with the "tkg" tag.

linux-tkg-config/prepare
damachine, line 772: And what do you think, maybe doing it this way makes sense here, too?

```bash
else
  _vanilla_tag=""
  _vanilla_mode && _vanilla_tag="vanilla-"
  echo "-$pkgrel-tkg-${_vanilla_tag}${_cpusched}${_compiler_name}" > localversion.10-pkgrel
  echo -e "Version tail set to \"-$pkgrel-tkg-${_vanilla_tag}${_cpusched}${_compiler_name}\"\n" > "$_where"/logs/prepare.log.txt
fi
```

PKGBUILD
```diff
@@ -55,10 +55,12 @@ fi
 
 source "$_where"/BIG_UGLY_FROGMINER
 
+_vanilla_tag=""
+_vanilla_mode && _vanilla_tag="vanilla-"
 if [ -n "$_custom_pkgbase" ]; then
   pkgbase="${_custom_pkgbase}"
 else
-  pkgbase=linux"${_basever}"-tkg-"${_cpusched}"${_compiler_name}
+  pkgbase="linux${_basever}-tkg-${_vanilla_tag}${_cpusched}${_compiler_name}"
```
damachine, line 32: Hi Adel the _vanilla_mode is defined in prepare.
Previously it was only sourced inside the if-block on first run.
On subsequent invocations (e.g. fakeroot package() phase),
BIG_UGLY_FROGMINER already exists, the if-block is skipped,
prepare is never sourced, and _vanilla_mode is undefined. :thinking: 

should we "source prepare" after BIG_UGLY_FROGMINER unconditionally to ensure all functions are always available?

my build log:

```
  BUILD   arch/x86/boot/bzImage
Kernel: arch/x86/boot/bzImage is ready  (#1)
==> Entering fakeroot environment...
           .---.`               `.---.
        `/syhhhyso-           -osyhhhys/`
       .syNMdhNNhss/``.---.``/sshNNhdMNys.
       +sdMh.`+MNsssssssssssssssNM+`.hMds+
       :syNNdhNNhssssssssssssssshNNhdNNys:
        /ssyhhhysssssssssssssssssyhhhyss/
        .ossssssssssssssssssssssssssssso.
       :sssssssssssssssssssssssssssssssss:
      /sssssssssssssssssssssssssssssssssss/   Linux-tkg
     :sssssssssssssoosssssssoosssssssssssss:        kernels
     osssssssssssssoosssssssoossssssssssssso
     osssssssssssyyyyhhhhhhhyyyyssssssssssso
     /yyyyyyhhdmmmmNNNNNNNNNNNmmmmdhhyyyyyy/
      smmmNNNNNNNNNNNNNNNNNNNNNNNNNNNNNmmms
       /dNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNd/
        `:sdNNNNNNNNNNNNNNNNNNNNNNNNNds:`
           `-+shdNNNNNNNNNNNNNNNdhs+-`
                 `.-:///////:-.`
/home/damachine/Projekte/linux-tkg/PKGBUILD: line 59: _vanilla_mode: command not found
chmod: cannot access '/tmp/makepkg/linux70-tkg-eevdf-llvm/pkg': No such file or directory
/usr/bin/fakeroot: line 178: 285623 User defined signal 1      FAKEROOTKEY=$FAKEROOTKEY LD_LIBRARY_PATH="$PATHS" LD_PRELOAD="$FAKEROOT_LIB" "$@"
  -> exit cleanup done
```

damachine (commented):
Hi Adel and all, hope everything is going well so far. So...

linux-tkg-config/prepare
AdelKS: Good catch! thanks!

PKGBUILD
AdelKS: good catch! thanks!

AdelKS:
Turns out user patches were actually getting applied nonetheless as they weren't going through the `_tkgpatcher` route. Just added the warning.

damachine:
~lol sry <3 no rush, maybe the message https://github.com/Frogging-Family/linux-tkg/pull/1204#discussion_r3249169185 is a bit hidden here 😄 What do you think about it?~

damachine:
ahh sorry, I saw it, very nice, a watermark, cool idea :rocket:

AdelKS:
> ahh sorry, I saw it, very nice, a watermark, cool idea 🚀

That's what it actually is, I renamed it but the function remains the same hehe

linux-tkg-config/prepare
damachine: Nothing serious, but it looks like an underline is missing here at `_f##` ;)

I was wondering whether we should extend the logging a little further.
If you activate the vanilla mod and tinker a bit with user patches, additional logging could help with analysis and troubleshooting.

Maybe you'll find something useful here:
https://github.com/damachine/linux-tkg/commit/97c18d110f0f8adfcadd498580c7dbb76be675b3#diff-5ee5543a04159ed0234b96931a4a912d6569924ed930fb0f75f1b2831d432f6cR632-R722

Tk-Glitch:
Alright, LGTM guys!
Thank you Adel, thank you damachine :+1:

PR #1210, opened 2026-05-23
prepare: harden exit cleanup

damachine:
Hi all <3, I looked into a small issues nothing wild

I suspect remove_deps is mostly a CI/makepkg-context thing. I have not fully validated every path yet, but we can safely avoid noisy command-not-found output when it is unavailable.

fix:
The logs directory now gets created before any cp/mv writes. Without that, files can land in the wrong place depending on execution context.

added:
A snapshot of the external cfg is now saved into logs as well, which should help with debugging and post-mortem analysis.

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1969,6 +1969,7 @@ exit_cleanup() {
   rm -f "$_where"/*.hook
   rm -f "$_where"/cleanup
   rm -f "$_where"/prepare
+  rm -f "$_where"/minimal-modprobed.db
 
   # Remove state tracker
   rm -f "$_where"/BIG_UGLY_FROGMINER
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -535,7 +535,7 @@ _setup_kernel_work_folder() {
   cd "$_where"
 
   if ! [ -d "$_kernel_source_folder_abs" ]; then
-    msg2 "First initialization of the linux source code git folder"
+    msg2 "First initialization of the linux source code git folder"  | tee -a "$_where/logs/git_work.log.txt"
     mkdir -p "$_kernel_source_folder_abs"
     cd "$_kernel_source_folder_abs"
     git -c init.defaultBranch=master init --bare
@@ -544,29 +544,30 @@ _setup_kernel_work_folder() {
   cd "$_kernel_source_folder_abs"
 
   if [[ ! "$_offline" =~ ^(y|Y|yes|true|True)$ ]]; then
-    msg2 "Fetching tag: $_kernel_git_tag from remote $_git_mirror" | tee "$_where/git_work.log.txt"
-    git fetch --depth 1 $_git_mirror tag "$_kernel_git_tag" &>> "$_where/git_work.log.txt"
+    msg2 "Fetching tag: $_kernel_git_tag from remote $_git_mirror" | tee -a "$_where/logs/git_work.log.txt"
+    git fetch --depth 1 $_git_mirror tag "$_kernel_git_tag" &>> "$_where/logs/git_work.log.txt"
   fi
 
-  msg2 "Checking out tag: $_kernel_git_tag" | tee -a "$_where/git_work.log"
-  msg2 "   in the work folder: $_kernel_work_folder_abs" | tee -a "$_where/git_work.log.txt"
+  msg2 "Checking out tag: $_kernel_git_tag" | tee -a "$_where/logs/git_work.log.txt"
+  msg2 "   in the work folder: $_kernel_work_folder_abs" | tee -a "$_where/logs/git_work.log.txt"
 
   # The space ' ' in grep -w "$_kernel_work_folder_abs " is important
   # to not match an existing folder with a longer name with the same prefix name
   if [ -d "$_kernel_work_folder_abs" ] && \
      ( git worktree list | grep -w "$_kernel_work_folder_abs " &> /dev/null ) && \
      ( cd "$_kernel_work_folder_abs" && git status &> /dev/null ); then
-    # Worktree folder exists and is a valid worktree
+    msg2 "Worktree folder $_kernel_work_folder_abs exists and is a valid worktree folder" | tee -a "$_where/logs/git_work.log.txt"
     cd "$_kernel_work_folder_abs"
-    git reset --hard &>> "$_where/git_work.log.txt"
-    git clean -ffdx &>> "$_where/git_work.log.txt"
-    git am --abort &>> "$_where/git_work.log.txt" || true
-    git rebase --abort &>> "$_where/git_work.log.txt" || true
-    git checkout "$_kernel_git_tag" &>> "$_where/git_work.log.txt"
+    git reset --hard &>> "$_where/logs/git_work.log.txt"
+    git clean -ffdx &>> "$_where/logs/git_work.log.txt"
+    git am --abort &>> "$_where/logs/git_work.log.txt" || true
+    git rebase --abort &>> "$_where/logs/git_work.log.txt" || true
+    git checkout "$_kernel_git_tag" &>> "$_where/logs/git_work.log.txt"
   else
     # In all other cases, just force create the work tree
-    rm -rf "$_kernel_work_folder_abs" &>> "$_where/git_work.log.txt"
-    git worktree add -f "$_kernel_work_folder_abs" "$_kernel_git_tag" &>> "$_where/git_work.log.txt"
+    msg2 "Force creating fresh git worktree folder $_kernel_work_folder_abs" | tee -a "$_where/logs/git_work.log.txt"
+    rm -rf "$_kernel_work_folder_abs" &>> "$_where/logs/git_work.log.txt"
+    git worktree add -f "$_kernel_work_folder_abs" "$_kernel_git_tag" &>> "$_where/logs/git_work.log.txt"
   fi
 
   _set_kver_internal_vars
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -2003,7 +2003,11 @@ exit_cleanup() {
   fi
 
   if [ "${_distro}" = "Arch" ]; then
-    remove_deps || ( true && msg2 "remove_deps not found" )
+    if command -v remove_deps &>/dev/null; then
+      remove_deps
+    else
+      msg2 "remove_deps not found"
+    fi
   fi
 
   msg2 'exit cleanup done\n'
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -573,6 +573,39 @@ _setup_kernel_work_folder() {
 }
 
 
+# Aggregate user configuration into BIG_UGLY_FROGMINER.
+# Duplicated variables are merged: env vars > ext config file > repo's customization.cfg
+# Strips blank lines and comments
+aggregate_user_config() {
+  local _cfg="${_where}/customization.cfg"
+
+  # Resolve _EXT_CONFIG_PATH from customization.cfg if not already in env
+  [[ -z "${_EXT_CONFIG_PATH:-}" ]] && eval "$(grep -m1 '^_EXT_CONFIG_PATH=' "$_cfg")"
+
+  [[ -f "${_EXT_CONFIG_PATH}" ]] && \
+    msg2 "External configuration file ${_EXT_CONFIG_PATH} will be used and will override customization.cfg values."
+
+  # Collect valid cfg keys to filter env overrides against
+  local -A _keys=()
+  local _k _v
+  while IFS='= ' read -r _k _; do
+    [[ -n "$_k" && "$_k" != '#'* ]] && _keys["$_k"]=1
+  done < "$_cfg"
+
+  # Emit env > ext > cfg, then strip blanks/comments and dedup by key (first wins)
+  {
+    while IFS='=' read -r _k _v; do
+      [[ -v "_keys[$_k]" ]] && printf '%s=%s\n' "$_k" "$_v"
+    done < <(printenv)
+
+    [[ -f "${_EXT_CONFIG_PATH}" ]] && cat "${_EXT_CONFIG_PATH}"
+    cat "$_cfg"
+  } | grep -v -E '^[[:space:]]*(#|$)' \
+    | awk -F= '!seen[$1]++' \
+    > "${_where}/BIG_UGLY_FROGMINER"
+}
+
+
 _tkg_initscript() {
 
   if ! whereis git > /dev/null 2>&1; then
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1981,60 +1981,6 @@ exit_cleanup() {
     rm -f "$_where"/linux"$_basever"-tkg-userpatches/"$_p"
   done
 
-  if [ "$_NUKR" = "true" ] && [ "$_where" != "$srcdir" ]; then
-    rm -rf "$_where"/src/*
-    # Double tap
-    rm -rf "$srcdir"/linux-*
-    rm -rf "$srcdir"/*.xz
-    rm -rf "$srcdir"/*.patch
-    rm -rf "$srcdir"/*-profile.cfg
-    rm -rf "$srcdir"/*.rules
-    rm -f "$srcdir"/config.x86_64
-    rm -f "$srcdir"/customization.cfg
-
-  elif [ "$_distro" == "Arch" ]; then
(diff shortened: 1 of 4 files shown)


linux-tkg-config/prepare
```diff
@@ -2003,20 +1996,36 @@ exit_cleanup() {
   fi
 
   if [ "${_distro}" = "Arch" ]; then
-    remove_deps || ( true && msg2 "remove_deps not found" )
+    command -v remove_deps &>/dev/null && remove_deps || ( true && msg2 "remove_deps not found" )
```
AdelKS: `remove_deps` doesn't exist in PKGBUILD so not sure what the heck this line is even doing here, let's just drop it altogether. Unless @Tk-Glitch knows the special magic

linux-tkg-config/prepare
```diff
@@ -2003,20 +1996,36 @@ exit_cleanup() {
...
-  cp -f "$_where"/customization.cfg "$_where"/logs/customization.cfg.txt
+  # ensure logs dir exists
+  mkdir -p "${_where}/logs"
+
+  # snapshot customization.cfg
+  [[ -f "${_where}/customization.cfg" ]] && \
+    cp -f "${_where}/customization.cfg" "${_where}/logs/customization.cfg.txt"
+
+  # snapshot external config if present
+  if [[ -n "${_EXT_CONFIG_PATH:-}" && -f "${_EXT_CONFIG_PATH}" ]]; then
```
AdelKS: I think we can just go with `-f "${_EXT_CONFIG_PATH}"` and if it's empty, it's not a file anyway

damachine: Yeah, you're right. The "-n" is redundant

damachine: okidoki should be right now, thanks

linux-tkg-config/prepare
AdelKS: what does it say on your machine, `remove_deps not found` ? If it is the case, let's go ahead and simply remove this whole if block. I guess we can revert it if ever TkG chimes in.

damachine: hehe i see where it come from.

remove_deps is a funtion in /usr/bin/makepkg 
Bildschirmfoto_20260524_133328


> what does it say on your machine, `remove_deps not found` 

I noticed it when I play with the `./patchwork` script

AdelKS: aha!

AdelKS: Then please go for a clean version: if the command exists, run it, and only otherwise say it doesn't exist. with `if` `then` `else` x)

damachine: 👍

AdelKS:
Testing it on my machine then we're good to go!

linux-tkg-config/prepare
```diff
@@ -2003,20 +1996,39 @@ exit_cleanup() {
...
+  # ensure logs dir exists
+  mkdir -p "${_where}/logs"
+
+  # snapshot customization.cfg
+  [[ -f "${_where}/customization.cfg" ]] && \
+    cp -f "${_where}/customization.cfg" "${_where}/logs/customization.cfg.txt"
+
+  # snapshot external config if present
+  [[ -f "${_EXT_CONFIG_PATH}" ]] && \
+    cp -f "${_EXT_CONFIG_PATH}" "${_where}/logs/ext_customization.cfg.txt"
```
AdelKS: Let's have a Quality of Life (QoL) addition ! If you are willing to:

- Remove comments and blank lines from the customization files while copying them
- Merge the two customization files and remove duplicates according to priorities: ext entries take priority

damachine: That sounds good. Hehe, do you have a specific suggestion for how we can implement this as simply as possible? x)

AdelKS: Yes I do ! I didn't want the steal the fun from you until you ask

- Remove blank lines and comments
  - Regex for blank lines `^$` (start of line and end of line are next to each other)
  - Regex for comment lines `^#` (`#` comes right after start of line)
  - so we do `cat customization.cfg | grep -v -e "^$" -e "^#"`
  - It doesn't catch _everything_ but KISS

```shell
# So we do that first for the ext config file then for the repo file
cat ext_file.cfg | grep -v -e "^$" -e "^#" > customization.cfg.tmp
cat repo_file.cfg | grep -v -e "^$" -e "^#" >> customization.cfg.tmp
```

then we stable sort and uniq (which keeps the first occurence)
```shell
cat customization.cfg.tmp | sort -s | uniq >  customization.cfg.txt
rm customization.cfg.tmp
```

That should be it 🤔

AdelKS: You can test this by simply running `./install.sh config` (it will do the cleanup and logs without building anything)

damachine: Thanks a lot. I'll take a look to ensure it stays KISS. :P

damachine: Nice idea the log looks nicer now. thanks. Let's see what you mean.

AdelKS: Okay I was stupid, that `uniq` idea isn't quite it since the lines start the same but are not really the same haha, so even stable sort actually fails.

Since I don't know all the tricks, Claude introduced me to some nice things with `awk` (it's always `awk`... Never got the chance to learn everything it can do)

Here's the block
```shell
{
  [[ -f "${_EXT_CONFIG_PATH}" ]] && \
    grep -v -e "^$" -e "^#" "${_EXT_CONFIG_PATH}"
  [[ -f "${_where}/customization.cfg" ]] && \
    grep -v -e "^$" -e "^#" "${_where}/customization.cfg"
} | awk -F= '!seen[$1]++' > "${_where}/logs/customization.cfg.txt"
```

I will copy the explanation verbatim

> #### `awk -F= '!seen[$1]++'`
>
> `-F=`
>
> Sets the **field separator** to `=`. For a line like `_distro="Gentoo"`, awk splits it into:
>
> | field | value |
> |-------|-------|
> | `$1`  | `_distro` |
> | `$2`  | `"Gentoo"` |
> | `$0`  | the whole line |
>
> Only the *first* `=` matters for `$1`; further `=` just create more fields (`$3`, `$4`, …), which we don't care about. That's why `KCFLAGS="-march=native"` is safe — `$1` is still `KCFLAGS`.
>
> #### `!seen[$1]++`
>
> This is one of awk's classic idioms. Three things happen at once:
>
> 1. **`seen[$1]`** — `seen` is an associative array (created on first use, no declaration needed), indexed by the variable name from the current line. Unset entries read as `0`.
> 2. **`seen[$1]++`** — *post*-increment. The expression evaluates to the **current** value, then bumps the stored value by 1.
>  - First time we see `_distro`: expression is `0`, array becomes `1`.
>   - Second time: expression is `1`, array becomes `2`.
> 3. **`!seen[$1]++`** — negation. `!0` is true (`1`), `!1` (or any non-zero) is false (`0`).
>
> #### Program structure
>
> A bare expression with no action block is shorthand for: `{ if (expr) print $0 }`
>
> So lines where the expression is true get printed.
>
> #### Net effect
>
> | occurrence of `$1` | `seen[$1]` before | `!seen[$1]++` | action     | `seen[$1]` after |
> |--------------------|-------------------|---------------|------------|------------------|
> | 1st                | `0`               | true          | **print**  | `1`              |
> | 2nd                | `1`               | false         | skip       | `2`              |
> | 3rd                | `2`               | false         | skip       | `3`              |
> 
> Keep the **first** occurrence of each variable name, drop the rest. Since EXT is fed in before the base config, EXT's value is the one that survives.

The more you know...

damachine: haha I also had some `awk seen` magic going on :P

> Since I don't know all the tricks,....

no worries, who knows all the tricks anyway — when in doubt, Claude it out lol

I had a similar idea but was thinking - what if we keep both values instead of deduplicating? So if ext overrides something, you'd still see the base/default value right above it as a reference. Could be nice for debugging, like "oh so that was the default".

Would something like that make sense or do you prefer the clean single-value output?

damachine: ah ok ok i understand now and re-checked and test it and my previous sort/uniq idea was wrong for this use case.

So awk is the correct approach here!

Because ext config is streamed first, ext values win on duplicate keys.

Thanks for catching that.

AdelKS: > Would something like that make sense or do you prefer the clean single-value output?

Well we just want to know what settings the user did, and actually this is not enough, because there's a way to override even the ext config file: env vars... haha

damachine: I understand that. I am with you. :+1: An idea? hehe

AdelKS: but the env vars are saved in BIG_UGLY_FROGMINER, so we can check which variables there are the same as in the initial repo customization and keep those, and then append them even above ext config. then we truly have everything.

damachine: I’ve made a few adjustments. Let's see if I understood you correctly. see ya. KISS.

AdelKS: That reminded me that we do something similar in `PKGBUILD` lines `29-54`, where we basically aggregate user config, with correct priorities, but without cleaning up comments / white spaces / unrelated user env vars.

So ! haha... Next step, move` PKGBUILD` line `29-54` into `prepare` as `aggregate_user_config` which creates a lean BIG_UGLY_FROGMINER, and we just move that as `logs/customization.cfg.txt` at cleanup time. We can make `install.sh` use it too instead of some parallel code that does the same too (lines `40-54`.)

It will have a few more intermediary variables, and they will be useful for debugging actually. Whadyathink ?

Sorry for the ping pong, interaction gives ideas. If you have had enough, we can merge it like this and I'll do the next step :P

damachine: I took a look at it and yea that's actually a really clean idea. centralizing it in `aggregate_user_config` and having one lean `BIG_UGLY_FROGMINER` as the single source of truth makes total sense. no more parallel code doing the same thing in different places 👌

> It will have a few more intermediary variables, and they will be useful for debugging actually. Whadyathink ?

Let's go! since nvidia-all probably has the same thing I'm actually curious now. I would take a look at what you mean. no need to go out of your way though, if you already have something in mind just say the word and we'll do it. just so I can follow along and understand what's going on xD

i am not sure if you can directly make changes in code? would be cool to see ha

> Sorry for the ping pong,....

haha no worries at all, ping pong is fine by me :P honestly love that the ping pong sparks new ideas for you, that's what makes it fun ;) you're thorough, you go at this the right way and I dig that - really fun working with you, thanks for putting in the time and effort!

damachine:
Sorry, got one more ball to play your way, hope I'm not being annoying with the ping pong. No rush, main thing is we move it forward the right way x)

This shouldn't be a big deal, but I have to ask if: `declare -p -x >> "$_where"/BIG_UGLY_FROGMINER` stores ALL exported shell variables from the current session. This may include secrets. because the file is located in the Git tracked directory ($PWD). If exit_cleanup() doesn't work (e.g. PKGBUILD crashes before the prepare script), the file stays there. A `git add .` would commit it accidentally. An .gitignore entry BIG_UGLY_FROGMINER as a easy fix? Whadyathink, am I too paranoid at this point x)?

Aha, I see that the `current_env` equivalent for `install.sh` is included in the .gitignore.

AdelKS:
> declare -p -x >> "$_where"/BIG_UGLY_FROGMINER 

You make the dictionary of actually used variables from the repo's customization file first (which you wrote already),  and filter the output of `declare -p -x` with it before it lands in BIG_UGLY_FROGMINER. I don't remember exactly but maybe `declare -p -x` has a weird output that isn't only `var=foo` but instead something like `declare -x var=foo` per line that we need to find a solution to.

damachine:
>  I don't remember....
:100: :+1: 
 
Okay, I've pushed some initial changes for the `aggregate_user_config` feature in a second commit. My first test with it produced the desired result, assuming I've understood everything correctly: 

priority env var > ext config > upstream customization.cfg, only known upstream keys kept, comments and duplicates stripped.

AdelKS:
It looks like `env` or `printenv` works better as it doesn't have those spurious `declare -x`, tested it on Termux (android) on bash and zsh, I'll test back at home. That would make the code simpler.

AdelKS:
Did a little pass on top of your commit, tell me what you think :P

linux-tkg-config/prepare
```diff
@@ -2003,20 +2026,38 @@ exit_cleanup() {
   fi
```
AdelKS: I don't think this is useful anymore since we moved to using `git` for managing the sources and simply doing `git reset --hard && git clean -ffdx`  in `_setup_kernel_work_folder`.

damachine: Brilliant!
Yeah, that would be really cool. If you're still totally into it: Go for it feel free to keep pushing it forward! Just don't let me hold you up! :P My battery is dead for today! xD <3

AdelKS: done, need you to do a quick test on Arch to see that I didn't break anything: check that you can start a `makepkg -si` right after an aborted one, with both `_NUKR=true` and `_NUKR=false`

damachine: upps `_NUKR=true` right after an aborted one

```
...
==> Starting prepare()...
ln: failed to create symbolic link '/tmp/makepkg/linux-tkg-vanilla/src/linux-src-git': File exists
==> ERROR: A failure occurred in prepare().
    Aborting...
  -> exit cleanup done
...
```

change here `ln -s` with force to `ln -sf` should be enough 

https://github.com/damachine/linux-tkg/blob/363e6fdd0f58498575589855e8cf4de0d74900ff/PKGBUILD#L99

with this change both `_NUKR=true` and `_NUKR=false` runs well. right after an aborted one

AdelKS: that's exactly what I had in mind, thanks!

damachine:
Hehe. Is it intentional that `minimal.db` and `new kernel.config` remain in the `PWD` dir after cleanup (for reuse, modification, or customization)? Or do they get overwritten during a re-run? If so, they could also be included in the cleanup step. :P

Otherwise, everything looks good. It runs exactly as desired. See ya x)

AdelKS:
`kernelconfig.new` staying there is intentional, but I guess could be moved to the logs too, but lets just leave it here for now: it's for us to know what's the final config used when building the kernel, without having to boot it.

Yeah that `minimal.db` was never being cleaned up (but is overwritten), but just an omission. I think we're good to go! Thanks! :frog:

linux-tkg-config/prepare
```diff
@@ -548,7 +548,7 @@ _setup_kernel_work_folder() {
     git fetch --depth 1 $_git_mirror tag "$_kernel_git_tag" &>> "$_where/git_work.log.txt"
   fi
 
-  msg2 "Checking out tag: $_kernel_git_tag" | tee -a "$_where/git_work.log"
+  msg2 "Checking out tag: $_kernel_git_tag" | tee -a "$_where/git_work.log.txt"
```
AdelKS: Why this change ? because we write everything with `.log` in `$_where` and only when we cleanup that we append the `.txt` while moving stuff to the `logs` folder.

we append the `.txt` so people can drag and drop them in the issues here

AdelKS: Oh I see, you fixed a typo, but at "wrong place"

AdelKS: actually, not the "wrong place". Let me think

damachine: Yep, we already use `.log.txt` everywhere, but this creates a lonely `git_work.log` in the repo root with only one info line inside :)

This way, only a single line I needed to change.

AdelKS:
I cleaned up the very first commit as it was introducing code that we'd later remove / edit / move

linux-tkg-config/prepare
AdelKS: I made the code that creates `git_work.log.txt` put directly in the `logs` folder. which gets re-created in the beginning of the script actually.

damachine:
> I made the code that creates git_work.log.txt put directly in the logs folder. which gets re-created in the beginning of the script actually.

I like the idea of ​​writing directly to the logs.

> I cleaned up the very first commit as it was introducing code that we'd later remove / edit / move

Absolutely np :+1: 

I see you move also `BIG_UGLY_FROGMINER` stats to logs. Nice!!

So, all in all, lgtm!

AdelKS:
> I see you move also BIG_UGLY_FROGMINER stats to logs. Nice!!

Yeah, that ended up being the main addition of this PR: unify everything into BIG_UGLY_FROGMINER and use that as final state that is useful in logs. 

Originally you wanted to also copy the external config, because we were copying the `customization.cfg` file into the logs.

I see that you are now signing your commits with your real name, I edited the commits to reflect that.

I think we are good to go this time, but maybe more marination time will help uncover an even better approach somewhere haha

Issue #1161, opened 2025-11-07
Fedora 43, Problem: problem with installed package


Disruptive-Change:
### Same-issue check

Yes

### Linux Distribution

Fedora

### Kernel version

6.17.7_tkg_bore

### What happened?

How may i fix this ?

RPM build warnings:
    absolute symlink: /lib/modules/6.17.7_tkg_bore/build -> /usr/src/kernels/6.17.7_tkg_bore
 -> Building successfully finished!
Do you want to install the new Kernel ? Y/[n]: Y
[sudo] password for osi-don: 
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: problem with installed package
  - installed package kernel-devel-matched-6.17.7-300.fc43.x86_64 requires kernel-devel = 6.17.7-300.fc43, but none of the providers can be installed
  - package kernel-devel-matched-6.17.7-300.fc43.x86_64 from updates requires kernel-devel = 6.17.7-300.fc43, but none of the providers can be installed
  - package kernel-devel-matched-6.17.1-300.fc43.x86_64 from fedora requires kernel-devel = 6.17.1-300.fc43, but none of the providers can be installed
  - cannot install both kernel-devel-6.17.7_tkg_bore-1.x86_64 from @commandline and kernel-devel-6.17.7-300.fc43.x86_64 from @System
  - cannot install both kernel-devel-6.17.7_tkg_bore-1.x86_64 from @commandline and kernel-devel-6.17.7-300.fc43.x86_64 from updates
  - cannot install both kernel-devel-6.17.7_tkg_bore-1.x86_64 from @commandline and kernel-devel-6.17.1-300.fc43.x86_64 from fedora
  - conflicting requests
You can try to add to command line:
  --skip-broken to skip uninstallable packages
 -> exit cleanup done

### Relevant logs

[customization.cfg.txt](https://github.com/user-attachments/files/23425259/customization.cfg.txt)
[git_work.log.txt](https://github.com/user-attachments/files/23425260/git_work.log.txt)
[prepare.log.txt](https://github.com/user-attachments/files/23425258/prepare.log.txt)
[shell-output.log.txt](https://github.com/user-attachments/files/23425256/shell-output.log.txt)
[system-info.txt](https://github.com/user-attachments/files/23425261/system-info.txt)

### My Favorite type(s) of frog

Frog

aamaanaa:
[...]

PR #1213, opened 2026-06-04
patch(6.18 7.0 7.1): refresh 0013-fedora-rpm.patch

damachine:
Add the standard Fedora kernel-devel Provides to the linux-tkg devel RPM.

This makes the package visible as a matching kernel-devel provider for module
build tooling and marks it as an installonly kernel package, allowing linux-tkg
kernel-devel to be installed alongside Fedora kernel-devel instead of causing
DNF transaction conflicts with kernel-devel-matched.

Should solve #1161 

fyi the fedora patches were already solved the same way in 6.1 and 6.6, nothing new really 🐸

### DIFF
--- a/linux-tkg-patches/6.18/0013-fedora-rpm.patch
+++ b/linux-tkg-patches/6.18/0013-fedora-rpm.patch
@@ -4,8 +4,8 @@ Date: Sat, 7 Feb 2026 18:52:48 +0100
 Subject: [PATCH] package: kernel.spec: various RPM fixes
 
 ---
- scripts/package/kernel.spec | 19 ++++++++++++++++++-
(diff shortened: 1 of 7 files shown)


damachine:
ahh I forget `./patchwork` crashes when the `logs/` directory doesn't exist yet. Added a `mkdir` guard for now - try running patchwork without the logs dir and you'll see what I mean:

```
tee: logs/git_work.log.txt: No such file or directory
mv: cannot move 'BIG_UGLY_FROGMINER' to 'logs/customization.cfg.txt': No such file or directory
```

Not sure if that's the right fix or if you have a better approach. xD see ya <3

AdelKS:
I am happy to see that the fix isn't anything special, thanks!

PR #1218, opened 2026-06-26
Centralize frog banner output and add _die helper

damachine:
Hi all :frog:!

I wanted to propose a few small nerd knobs for linux-tkg. xD

Nothing too crazy here, mostly some small cleanup plus one optional THP knob. Hope you like it

~- Adds `_hugepage`, allowing users to choose the default THB mode: `always`, `madvise`, or `never`. Defaults stay unchanged: THP is left untouched unless users set `_hugepage`.~

- And while I was already poking around in `prepare`, I added a small `_die()` helper and routed the fatal `exit 1` paths through it. Some of them were printing warnings or regular messages right before stopping the script, which felt a bit shy for something that ends the party. xD

- Tiny extra nerd polish: the frog banner now lives in one shared helper, and install.sh gets to enjoy it too.

Sorry for the suggestion pile. I tried a few small things here; keep what you like and toss the rest. See ya :P

KISS <3

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -5,28 +5,10 @@
 # Contributor: Tk-Glitch <ti3nou at gmail dot com>
 # Contributor: Hyper-KVM <hyperkvmx86 at gmail dot com>
 
-plain '       .---.`               `.---.'
-plain '    `/syhhhyso-           -osyhhhys/`'
-plain '   .syNMdhNNhss/``.---.``/sshNNhdMNys.'
-plain '   +sdMh.`+MNsssssssssssssssNM+`.hMds+'
-plain '   :syNNdhNNhssssssssssssssshNNhdNNys:'
-plain '    /ssyhhhysssssssssssssssssyhhhyss/'
-plain '    .ossssssssssssssssssssssssssssso.'
-plain '   :sssssssssssssssssssssssssssssssss:'
-plain '  /sssssssssssssssssssssssssssssssssss/   Linux-tkg'
-plain ' :sssssssssssssoosssssssoosssssssssssss:        kernels'
-plain ' osssssssssssssoosssssssoossssssssssssso'
-plain ' osssssssssssyyyyhhhhhhhyyyyssssssssssso'
-plain ' /yyyyyyhhdmmmmNNNNNNNNNNNmmmmdhhyyyyyy/'
-plain '  smmmNNNNNNNNNNNNNNNNNNNNNNNNNNNNNmmms'
-plain '   /dNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNd/'
-plain '    `:sdNNNNNNNNNNNNNNNNNNNNNNNNNds:`'
-plain '       `-+shdNNNNNNNNNNNNNNNdhs+-`'
-plain '             `.-:///////:-.`'
-
 _where="$PWD" # track basedir as different Arch based distros are moving srcdir around
 
 source "$_where"/linux-tkg-config/prepare
+_frog_banner
 
 # Create BIG_UGLY_FROGMINER only on first run and save in it all settings
 if [ ! -e "$_where"/BIG_UGLY_FROGMINER ]; then
--- a/install.sh
+++ b/install.sh
@@ -162,7 +162,12 @@ if [ "$1" = "install" ]; then
       _kernel_flavor="tkg-${_cpusched}${_compiler_name}"
     fi
   else
-    _kernel_flavor="tkg-${_kernel_localversion}"
+    if [[ "$_kernel_localversion" != *tkg* ]]; then
+      _localversion_tag="tkg-"
+    else
+      _localversion_tag=""
+    fi
+    _kernel_flavor="${_localversion_tag}${_kernel_localversion}"
   fi
 
   # Setup kernel_subver variable
--- a/install.sh
+++ b/install.sh
@@ -47,7 +47,9 @@ if which script &> /dev/null && [[ "$_logging_use_script" =~ ^(Y|y|Yes|yes)$ && 
   exit
 fi
 
-source linux-tkg-config/prepare && aggregate_user_config
+source linux-tkg-config/prepare
+_frog_banner
+aggregate_user_config
 source "$_where/BIG_UGLY_FROGMINER"
 
 ####################################################################
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -117,6 +117,27 @@ _die() {
   exit 1
 }
 
+_frog_banner() {
+  plain '       .---.`               `.---.'
+  plain '    `/syhhhyso-           -osyhhhys/`'
+  plain '   .syNMdhNNhss/``.---.``/sshNNhdMNys.'
+  plain '   +sdMh.`+MNsssssssssssssssNM+`.hMds+'
+  plain '   :syNNdhNNhssssssssssssssshNNhdNNys:'
+  plain '    /ssyhhhysssssssssssssssssyhhhyss/'
+  plain '    .ossssssssssssssssssssssssssssso.'
+  plain '   :sssssssssssssssssssssssssssssssss:'
+  plain '  /sssssssssssssssssssssssssssssssssss/   Linux-tkg'
+  plain ' :sssssssssssssoosssssssoosssssssssssss:        kernels'
+  plain ' osssssssssssssoosssssssoossssssssssssso'
+  plain ' osssssssssssyyyyhhhhhhhyyyyssssssssssso'
+  plain ' /yyyyyyhhdmmmmNNNNNNNNNNNmmmmdhhyyyyyy/'
+  plain '  smmmNNNNNNNNNNNNNNNNNNNNNNNNNNNNNmmms'
+  plain '   /dNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNd/'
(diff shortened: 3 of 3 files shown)


AdelKS:
Hey,

Thanks for you help again ! I'd love to merge this except the hugepage knob, I don't think there's any real benefit to offer the knob as the default is good for everyone. [The documentation](https://www.kernel.org/doc/html/latest/admin-guide/mm/transhuge.html) seems to corroborate

> Transparent Hugepage Support for anonymous memory can be disabled (mostly for debugging purposes) or only enabled inside MADV_HUGEPAGE regions (to avoid the risk of consuming more memory resources) or enabled system wide

If you still really want it in, I'll let TkG make the final call :P

AdelKS:
Oh right, are you aware of the `.myfrag` approach here ? to have your own kind of config option changes on top what's done automatically or with knobs, personal changes to THP should fit there without any issues.

Tk-Glitch:
> Hey,
> 
> Thanks for you help again ! I'd love to merge this except the hugepage knob, I don't think there's any real benefit to offer the knob as the default is good for everyone. [The documentation](https://www.kernel.org/doc/html/latest/admin-guide/mm/transhuge.html) seems to corroborate
> 
> > Transparent Hugepage Support for anonymous memory can be disabled (mostly for debugging purposes) or only enabled inside MADV_HUGEPAGE regions (to avoid the risk of consuming more memory resources) or enabled system wide
> 
> If you still really want it in, I'll let TkG make the final call :P

I'd like to have damachine's point of view regarding why we should have this in. I don't think we *need* it, but I'm not strongly opposed to it if there's a justification I fail to see. Having too many options is as detrimental as not having enough imho, since those are directly aimed at ease of use.

Other than that, I'm fine merging.

damachine:
Hi guys, thanks for the feedback.

Yeah, fair point. I also see the danger of adding too many knobs. At some point the config turns into a spaceship cockpit. xD

Maybe I can explain my reasoning a bit better.

We already tune the THP defrag side with the zen `defer+madvise` patch here:
https://github.com/Frogging-Family/linux-tkg/blob/af36066ddbaa1d66071ab6aabcd19abd34201448/linux-tkg-patches/7.2/0003-glitched-base.patch#L221

But that is only about how reclaim/compaction should behave when THP is being used or when THP allocation falls back.

The knob I added is about the other side, the main THP policy itself (`always`, `madvise`, `never`).

So my thought was: if we already carry a patch to make the defrag behavior more suitable for interactive systems, it is not completely out of place to also expose the main THP policy as an optional linux-tkg knob.

It is mostly for users who want a reproducible config-side way to pick the THP policy instead of doing it later through `.myfrag`, sysfs/tmpfiles, boot parameters, or distro-specific tweaks.

But yeah, absolutely, I totally get the “too many knobs” concern. Maybe This is more of a convenience knob, not a must-have. If you think it adds more noise than value, I’m fine dropping the THP part and keeping only the helper bits.

AdelKS:
My current understanding is that there's no workload that benefits from disabling THP and it brings a net benefit: it allow the kernel to track RAM usage in chunks bigger than the default 4kB (that's what e.g. what makes segfaults work: accessing memory outside of the originally attributed 4kB chunk). I've learned recently that `glibc` allocator already uses `madvise` when e.g. doing `malloc()` / `new` with a big enough chunk: so potentially everything in the OS is using it. Disabling it just sounds counter-productive. The patch you linked made me understand that there can be scheduling spikes when the kernel defrags the THP pool of the given chunk size (there can be many) and that's what that patch is tweaking.

ptr1337:
> My current understanding is that there's no workload that benefits from disabling THP and it brings a net benefit: it allow the kernel to track RAM usage in chunks bigger than the default 4kB (that's what e.g. what makes segfaults work: accessing memory outside of the originally attributed 4kB chunk). I've learned recently that `glibc` allocator already uses `madvise` when e.g. doing `malloc()` / `new` with a big enough chunk: so potentially everything in the OS is using it. Disabling it just sounds counter-productive. The patch you linked made me understand that there can be scheduling spikes when the kernel defrags the THP pool of the given chunk size (there can be many) and that's what that patch is tweaking.

It can improve latency for some specific tasks, but well. Using Madvise is in general the "smartest" way to use, but could still have a loss in applications not requesting hugepages to be used. Madvise should have a bit lower latency then full too, but full has better throughput.

So in the end, it just depends on workload.

damachine:
> My current understanding is that there's no workload that benefits from disabling THP and it brings a net benefit: it allow the kernel to track RAM usage in chunks bigger than the default 4kB (that's what e.g. what makes segfaults work: accessing memory outside of the originally attributed 4kB chunk). I've learned recently that `glibc` allocator already uses `madvise` when e.g. doing `malloc()` / `new` with a big enough chunk: so potentially everything in the OS is using it. Disabling it just sounds counter-productive. The patch you linked made me understand that there can be scheduling spikes when the kernel defrags the THP pool of the given chunk size (there can be many) and that's what that patch is tweaking.

Yeah, fair point. Also, the detail was really helpful, thanks. :P

I didn’t mean to suggest that disabling THP is generally useful for normal desktop use. My idea was more about testing/comparison and some edge cases where someone deliberately wants to try `madvise` or `never`.

But yeah, that is probably too niche for a normal linux-tkg knob.

AdelKS:
Ptr is saying that it can have a noticeable overhead, is disabling it at runtime the same as not building it altogether in terms of latency improvement ? Because if not-built-at-all is better than disabled-at-runtime, then it can be a good knob :thinking:  If not, I am for leaving it out of the customization file.

damachine:
> Ptr is saying that it can have a noticeable overhead, is disabling it at runtime the same as not building it altogether in terms of latency improvement ? Because if not-built-at-all is better than disabled-at-runtime, then it can be a good knob 🤔 If not, I am for leaving it out of the customization file.

My knob only changes the main THP policy, so never would still mean THP is built but disabled through the normal policy path. It does not disable CONFIG_TRANSPARENT_HUGEPAGE at build time.

So if the interesting latency question is “not built at all” vs “built but set to never”, my current knob does not really cover that stronger case anyway. Not sure if there is a meaningful difference there.

Okidoki, I think these are the final questions now. xD

For always/madvise, it is more of a policy/convenience thing. We could maybe expose only those two as a rough throughput vs more conservative/latency-friendly user choice, but I also agree that kernel cmdline, sysfs/tmpfiles.d, or .myfrag can already cover it.

Whatever you prefer here is fine with me.

AdelKS:
> So if the interesting latency question is “not built at all” vs “built but set to never”, my current knob does not really cover that stronger case anyway. Not sure if there is a meaningful difference there.

I am well aware, the next thing if the answer is yes would be to change that knob into a "remove at build time" one, which is more worthwhile for a build-time customization option. 

> but I also agree that kernel cmdline, or sysfs/tmpfiles.d can already cover it.

Yeah exactly, so no need for one extra place to do the same. And customization isn't too good of a place since it would only change a default runtime behavior that can also be changed at runtime. (we probably already do that, like the default kernel cmdline knob, but I'd be reluctant to add more)

Bringing something into the `customization.cfg` does however have the effect of bringing the question into the mind of a `linux-tkg` user: I want that question to be worthwhile instead of time-wasting. I originally believed that the current default is good for 99.9% of us, now maybe not :thinking: But I still think it shouldn't be in `customization.cfg` because it's something that can be changed at runtime, and elsewhere already.

damachine:
Ookay, I understand you

I don’t think ptr said “noticeable overhead” directly. I understood it more as “policy can affect latency depending on workload”.

But yeah, your question is still the interesting one: is building without CONFIG_TRANSPARENT_HUGEPAGE meaningfully different.

Don't know if that makes sense.
I have never tested that. I can't tell you. :P

> Bringing something into the `customization.cfg`

Ah, sry, I get what you mean. we should really make a permanent change rather than something that can be change later on. I'll keep that in mind!! :+1: 

I'll drop the THB thing from the PR.

damachine:
Unfortunately, my knowledge of it is not deep enough but I looked around a bit. There are cases where disabling THP can make sense, mostly databases, in-memory services, cloud, server workloads....I don’t think that is really the main linux-tkg target audience. ;)

Dropping it feels like the sane option for now.

Thanks for now Adel for taking the time again with me.
That was a good point about not adding knobs for things that can already be changed at runtime. I’ll keep that in mind for future nerd ideas. :P

damachine:
>  I originally believed that the current default is good for 99.9% of us, now maybe not 🤔

Lol, sorry, that wasn't intentional xD I wouldn't worry too much about it.

if I’m reading this right :thinking:, we already inherit the Arch-style default (99,9%) here anyway, so THP should already default to `always` unless we override it somewhere later.

https://github.com/Frogging-Family/linux-tkg/blob/af36066ddbaa1d66071ab6aabcd19abd34201448/linux-tkg-config/7.1/config.x86_64#L1284

So yeah, sorry Adel bro, I think I was overthinking this one a bit. xD

Dropping the knob makes even more sense then.

damachine:
Okay, I hope this doesn't mess anything up, but I noticed one small annoying inconsistency.

I mirrored `prepare`'s `tkg` prefix handling for non-Arch custom localversions in `install.sh`. Arch uses `_custom_pkgbase` there, while `install.sh` uses `_kernel_localversion`.

So my suggestion is to only add the `tkg` prefix when the custom value does not already contain it.

AdelKS:
Alright let's go! thanks!

Issue #919, opened 2024-04-10
Arch: vm.max_map_count tweak superceded by config in filesystem-2024.04.07-1 or newer


thojohns:
As of April 7, 2024, Arch Linux has set its default value for `vm.max_map_count` to `1048576` [(source)](https://archlinux.org/news/increasing-the-default-vmmax_map_count-value/). This is higher than the previous default value of `65530` but lower than linux-tkg's default of `16777216`. However, Arch's method of enforcing this value - updating `/usr/lib/sysctl.d/10-arch.conf` in the core `filesystem` package to include the line `vm.max_map_count=1048576` - overrides any kernel-specific settings, unless a higher-priority config file sets it to another value.

The 'proper' solution would be to create that higher priority config file - create something like `/usr/local/lib/sysctl.d/99-vm-max-map-count` with the line `vm.max_map_count=16777216`. This is outside the scope of a kernel-only package (especially with the example `/usr/local` path), and probably useless for non-Arch distros, but it might be worth noting somewhere in the documentation anyways. I'm also uncertain if other distros that change the default `vm_max_map_count` value need similar overrides or not as I have no experience with most of them.

ryanmusante:
Will a sysctl.conf in /etc with this entry still take precedence?

thojohns:
> Will a sysctl.conf in /etc with this entry still take precedence?
sysctl.conf isn't even read from, at least on systemd-based Arch systems, and that appears to have been the case for the last decade; I can't really test on a non-systemd setup. 

The Arch wiki [recommends](https://wiki.archlinux.org/title/sysctl#Configuration) migrating `/etc/sysctl.conf` to `/etc/sysctl.d/99-sysctl.conf`, and I can at least confirm that dropping the example conf above into `/etc/sysctl.d/` also works - putting personal configs in `/usr/local/` is just a personal preference of mine.

PR #1221, opened 2026-07-04
Patch 7.1 7.2: Huge "summer" refresh

damachine:
This brings in a few newer Zen defaults for the new 7.x kernel line and drops some of the older, more conservative values we were carrying before. These are only suggestions of course, so please treat them as such.

Since 7.x is a new major kernel line, I thought this might be a good point to refresh the `glitched-base` defaults a little. The added Zen bits seem fairly long lived and well tested by now, so carrying them from 7.1/7.2 onward felt reasonable to me.

The maintenance cost should hopefully stay fairly low too, since these are mostly small default/value changes. Also, some of these changes only take effect when the related `_zenify` / `.cfg` options are enabled, so they should not affect everyone unconditionally.

Hope you like the direction. If not, I will only cry a very small, properly rate limited amount. xD

Imported from:

https://github.com/zen-kernel/zen-kernel/commit/a07ceee31a99bc8b29c07ee5183fa39e84b989aa

1. Refreshes `0013-optimize_harder_O3.patch` for 7.1 and 7.2.

I cherry-picked these bits from Zen and added them to O3. (Perhaps the choice of location in O3 isn't ideal; it might be better placed in base/misc.)

- removes GCC `-fconserve-stack` [0006-zen-disable-stack-conservation-for-gcc.patch](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0006-zen-disable-stack-conservation-for-gcc.patch)
- disables AVX2/tree vectorization on x86 with:`-mno-avx2 -fno-tree-vectorize` [0007-zen-arch-x86-disable-avx2-and-tree-vectorization.patch](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0007-zen-arch-x86-disable-avx2-and-tree-vectorization.patch)
- adds optional Clang Polly optimization for LLVM O3 builds when `LLVMPolly.so` is available https://github.com/CachyOS/kernel-patches/blob/master/7.1/misc/0001-clang-polly.patch

2. Refreshes `0003-glitched-base.patch` for 7.1 and 7.2.

New Zen patches added:

- [`mm: stop kswapd early when no allocation waiters remain`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0015-zen-mm-stop-kswapd-early-when-nothing-s-waiting-for-it-to-free-pages.patch)
- [`mm: disable watermark boosting by default for interactive builds`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0027-zen-interactive-mm-disable-watermark-boosting-by-default.patch)
- [`mm/swap: disable swap-in readahead`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0028-zen-interactive-mm-swap-disable-swap-in-readahead.patch)
- [`cpufreq: remove the schedutil dependency from Intel/AMD P-State drivers`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0012-zen-cpufreq-remove-schedutil-dependency-on-intel-amd-p-state-drivers.patch)
- [`x86: disable split lock mitigation by default for interactive builds`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0030-zen-interactive-disable-split-lock-mitigation-by-default.patch)
- [`sched/core: always inline finish_task_switch() and related helpers`](https://github.com/damachine/linux-tkg/blob/staging/linux-tkg-patches/7.1/patchwork/zen-split/0018-zen-sched-core-make-finish-task-switch-and-its-subfunctions-always-inline.patch)
~*I’m not entirely sure about it, though, since it involves changes to the scheduler. As I see it, it shouldn't affect the other schedulers, but I can't say for certain whether there might be a conflict.*~ 
*I tested that with BORE and didn't notice any problems.*

Value changes:

* `max_map_count`: `16777216` -> `INT_MAX - 5`

  Replaced the older two-step TKG `max_map_count` handling (`262144` -> `16777216`) with Zen's style default: `INT_MAX - MAPCOUNT_ELF_CORE_MARGIN`, effectively `2147483642`.

* File readahead: `2M` -> `256K`

  This follows the CachyOS/Zen-style reasoning of doubling the default `VM_READAHEAD_PAGES` from 128K to 256K, increasing throughput a bit without going so far that it hurts latency too much.

* SQ scheduler: `mq-deadline` -> `bfq`

* MQ scheduler: `none` -> `kyber`

* Watermark boost factor: `15000` / `1.5` -> `0` under `CONFIG_ZENIFY`

* Swap-in readahead / `page_cluster`: default auto value -> `0` under `CONFIG_ZENIFY`

* Split lock mitigation: `1` -> `0` under `CONFIG_ZENIFY`

Other patchset changes:

* Updated the Zenify help text to document the new defaults.
* Refreshed/rebased the patchsets for the newer kernel bases:

  * 7.1 checked on v7.1.3
  * 7.2 checked on v7.2-rc2
  
Peace :frog:

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -130,6 +130,10 @@ _libunwind_replace=""
 # Set to "0" to disable if desired, otherwise stick to the default behavior.
 _llvm_ias="1"
 
+# Enable Clang Polly loop optimizations. Default is "1" when using LLVM.
+# Requires the LLVMPolly.so plugin to be available to Clang.
+_llvm_polly="1"
+
 # Clang LTO mode, only available with the "llvm" compiler - options are "no", "full" or "thin".
 # ! This is currently experimental and might result in an unbootable kernel - Not recommended !
 # "no: do not enable LTO"
--- a/customization.cfg
+++ b/customization.cfg
@@ -231,7 +231,7 @@ _aggressive_ondemand="true"
 _tcp_cong_alg=""
 
 # You can pass a default set of kernel command line options here - example: "intel_pstate=passive nowatchdog amdgpu.ppfeaturemask=0xfffd7fff mitigations=off"
-_custom_commandline="intel_pstate=passive split_lock_detect=off"
+_custom_commandline="intel_pstate=passive"
 
 # Selection of Clearlinux patches
 _clear_patches="true"
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1386,6 +1386,17 @@ _tkg_srcprep() {
     fi
   fi
 
+  tkgpatch="$srcdir/0013-clang-polly.patch"
+  if [ -e "$tkgpatch" ]; then
+    if [[ "$_llvm_polly" =~ ^(true|yes|1)$ ]] && [ "$_compiler" = "llvm" ]; then
+      _msg="Patching Clang Polly optimization"
+      _tkg_patcher
+      ! _vanilla_mode && _enable "POLLY_CLANG"
+    elif [[ "$_llvm_polly" =~ ^(true|yes|1)$ ]] && [ "$_compiler" != "llvm" ]; then
+      warning "Clang Polly requires LLVM, skipping."
+    fi
+  fi
+
   # irq threading
   if [ "$_irq_threading" = "true" ]; then
     _enable "FORCE_IRQ_THREADING"
--- a/linux-tkg-patches/7.1/0003-glitched-base.patch
+++ b/linux-tkg-patches/7.1/0003-glitched-base.patch
@@ -1,42 +1,43 @@
-From f7f49141a5dbe9c99d78196b58c44307fb2e6be3 Mon Sep 17 00:00:00 2001
+From bc11a34a1a07178adc9d3df7a6c0889b88e0d5ae Mon Sep 17 00:00:00 2001
 From: Tk-Glitch <ti3nou@gmail.com>
 Date: Wed, 4 Jul 2018 04:30:08 +0200
-Subject: [PATCH 01/17] glitched
+Subject: [PATCH 01/21] glitched
 
 ---
  init/Makefile | 2 +-
- 1 file changed, 1 insertions(+), 1 deletions(-)
+ 1 file changed, 1 insertion(+), 1 deletion(-)
 
 diff --git a/init/Makefile b/init/Makefile
-index baf3ab8d9d49..854e32e6aec7 100755
+index d6f75d890..9fa3e5890 100644
 --- a/init/Makefile
 +++ b/init/Makefile
-@@ -19,7 +19,7 @@ else
-
+@@ -34,7 +34,7 @@ build-timestamp = $(or $(KBUILD_BUILD_TIMESTAMP), $(build-timestamp-auto))
+ 
  # Maximum length of UTS_VERSION is 64 chars
  filechk_uts_version = \
 -	utsver=$$(echo '$(pound)'"$(build-version)" $(smp-flag-y) $(preempt-flag-y) "$(build-timestamp)" | cut -b -64); \
 +	utsver=$$(echo '$(pound)'"$(build-version)" $(smp-flag-y) $(preempt-flag-y) "TKG" "$(build-timestamp)" | cut -b -64); \
  	echo '$(pound)'define UTS_VERSION \""$${utsver}"\"
-
+ 
  #
 -- 
-2.28.0
+2.55.0
 
 
-From c304f43d14e98d4bf1215fc10bc5012f554bdd8a Mon Sep 17 00:00:00 2001
+From 6c215b31de789f44a5871f1fd77070509cece3bb Mon Sep 17 00:00:00 2001
 From: Alexandre Frade <admfrade@gmail.com>
 Date: Mon, 29 Jan 2018 16:59:22 +0000
-Subject: [PATCH 02/17] dcache: cache_pressure = 50 decreases the rate at which
+Subject: [PATCH 02/21] dcache: cache_pressure = 50 decreases the rate at which
  VFS caches are reclaimed
 
 Signed-off-by: Alexandre Frade <admfrade@gmail.com>
 ---
- fs/dcache.c | 2 +-
- 1 file changed, 1 insertion(+), 1 deletion(-)
+ fs/dcache.c       | 2 +-
+ kernel/sched/rt.c | 4 ++--
+ 2 files changed, 3 insertions(+), 3 deletions(-)
 
 diff --git a/fs/dcache.c b/fs/dcache.c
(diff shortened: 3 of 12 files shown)


AdelKS:
@Tk-Glitch leaving this to you :P

damachine:
FYI:

I'm just testing to see if everything is correct, and I just noticed that the `vm.max_map_count` patch default is overridden on Arch by `/usr/lib/sysctl.d/10-arch.conf`, which sets it to `1048576`.

So the kernel-side default currently does not survive userspace sysctl loading on Arch. This should affects both the older `DEFAULT_MAX_MAP_COUNT = 16777216` logic and the newer `DEFAULT_MAX_MAP_COUNT = INT_MAX - MAPCOUNT_ELF_CORE_MARGIN` logic.

So maybe linux-tkg would need either a later `sysctl.d` drop-in if this should apply OOTB, or at least a README note for Arch users that `/usr/lib/sysctl.d/10-arch.conf` overrides `vm.max_map_count` to `1048576`.

---

Small side question, because I may be missing something obvious here:

Arch overrides `vm.max_map_count` at runtime via `/usr/lib/sysctl.d/10-arch.conf`, so I’m not sure why Arch does not just build the kernel with that value directly as the default, instead of setting it again from userspace later. 🤔

The Zen-style change makes this even funnier/confusing to me: Zen raises the kernel-side default, but on Arch it gets lowered again once userspace sysctl loading applies `/usr/lib/sysctl.d/10-arch.conf`.

So unless I am missing something, the raised kernel default only really remains effective on systems where that Arch sysctl file is not loaded, or where the distro also raises the runtime sysctl value.

Maybe I am overthinking it, but this made my brain do a small :frog: jump. xD

Tk-Glitch:
It used not to be the case, but yeah right now it's basically useless as is on Arch. It'll however still work on distros not setting it from userspace.
We *could* bypass it in userspace ourselves on Arch at least. There's an open issue about it already btw.

damachine:
> It used not to be the case, but yeah right now it's basically useless as is on Arch. It'll however still work on distros not setting it from userspace.

Thanks for confirming, that was exactly what confused me too.

> We *could* bypass it in userspace ourselves on Arch at least.

Okay, did I understand that right, would shipping a `sysctl.d` with linux-tkg be the sane way here? Something like `99-linux-tkg-max-map-count.conf`, so the linux-tkg value does not get Arch-bonked again by `10-arch.conf`. Or is there a cleaner way to handle this?

> There's an open issue about it already btw.

Do you maybe have the link to that issue?

~I can add my findings there too, maybe one more confused frog voice helps Arch take another look at the `sysctl.d` bonk. xD~

ahh hehe you mean #919 maybe? i thought that was more on Arch’s side than linux-tkg :P

damachine:
Howdy @Tk-Glitch I also added the latest [Arch](https://github.com/archlinux/linux/releases/tag/v7.1.3-arch2) fix to the misc patch. Not sure if you need anything else from me, if you’re okay with it as it is, or if you’re just busy right now. No rush at all, hope all is fine <3

It’s been running perfectly for me on both Arch and Gentoo for a while so far. No performance or stability issues noticed, and hehe, no smoke coming out of the box either ;)

AdelKS:
Where did you get the `POLLY_CLANG` changes ?

damachine:
> Where did you get the `POLLY_CLANG` changes ?

From cachyos

ptr1337 has the best approach. it checks whether Polly is installed, and if not, it doesn't do anything. I’ve seen this with others like XanMod or LQX too, but they handle it more strictly; this approach was better.

https://github.com/CachyOS/kernel-patches/blob/master/7.1/misc/0001-clang-polly.patch

AdelKS:
I think you should separate the patch and have it have a standalone commit, so we can see the addition.

damachine:
> I think you should separate the patch and have it have a standalone commit, so we can see the addition.

Sorry, you’re right

Split Polly into its own commit now. The enablement is in prepare, only for `LLVM O3` builds, and `Kconfig` checks for `LLVMPolly.so` first.

AdelKS:
I mean also keep the polly patch standalone outside of the O3 patch, it's actually just a feature addition :P it will help us sync that patch with upstream with ci for example and avoid double work

damachine:
Then let’s take the Cachy kernel patch 1:1 as a standalone eg. `0013-clang-polly.patch`.
Cachy doesn’t have a 7.2 copy yet, but the 7.1 series applies cleanly there too, so we can keep a checked 7.2 copy on our side. does that sound right?

AdelKS:
Yeah exactly, that's a better approach. I'll let TkG and Ptr chime in on whether we should add the polly feature: it looks relatively low maintenance as the flag names and where they get added should be stable (does it touch the same spots where we add `-march=native` ?). It probably doesn't have any real perf improvement,  but a +0.1% is always better than 0% :P

damachine:
> (does it touch the same spots where we add `-march=native` ?)

Not really. Polly adds a separate Kconfig option and guarded KBUILD_CFLAGS block; it does not touch the -march=native path. Maintenance should be minimal beyond keeping the Polly flags current.

> It probably doesn't have any real perf improvement, but a +0.1% is always better than 0% :P

Yeah, tiny gains add up xD

Tk-Glitch:
Whoopsy, I've merged Arch's 7.1.4 misc patches in misc-additions before seeing this was updated. Sorry damachine.

I'm fine with the standalone clang-polly patch approach. If anything goes south with it people should be able to turn it off "easily". Now about "should we add it", I guess that's in line with the 0.00001% better perf philosophy that many of our users adhere to :frog:
Worth it.

Tk-Glitch:
> ahh hehe you mean #919 maybe? i thought that was more on Arch’s side than linux-tkg :P

It definitely is, but should we address it on our end or is the current situation fine as is? I *suppose* it's fine, but some games like Star Citizen might still hit the current limit in some situations without the user knowing that's why it crashed.
What do you guys think?
Nonetheless this can be merged and we can think about that on the side.
Thanks damachine and Adel! :frog: :heart:

damachine:
Oh, and theoretically, we could also use the Polly patch for older versions, though I think it would suffice for 6.12 and 6.18.

> What do you guys think?

Well, I’ll take what I can get. I think my system deserves it. :P

But yeah, Adel and I talked about that before too. My takeaway from that discussion was that we should probably just leave it as is if that's how Arch handles it.

Though a note about this setting in the new README would be good. Otherwise Adel might want to update his [gaming guide](https://github.com/AdelKS/LinuxGamingGuide), since it's an excellent source for details like that. xD

PR #1222, opened 2026-07-07
prepare: use kmsg DRM panic screen for LLVM LTO

damachine:
Damn, I really don’t have enough to do, huh? :P

Small detail, but I think `kmsg` is the more useful fallback here.

The `qr_code` DRM panic screen needs Rust, and Rust is not supported with LLVM/LTO yet. Because of that, the kernel falls back to `user`:

```text
[    0.309020] Unsupported value for CONFIG_DRM_PANIC_SCREEN ('qr_code'), falling back to 'user'
```

Using `kmsg` for LLVM/LTO builds seems more helpful than the plain `user` screen. If the system really hangs or panics, the last kernel messages on the panic screen give at least some hint about what happened.

I only noticed this while going through `dmesg`, so yeah, tiny detail. But it avoids the warning and makes the panic screen a bit more useful if someone ever ends up there.

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -1050,6 +1050,13 @@ _tkg_srcprep() {
       _disable LTO_CLANG_THIN
       _enable LTO_NONE
     fi
+
+    # QR panic screen needs Rust; use kmsg for llvm/lto builds
+    if [[ "$_lto_mode" == "thin" || "$_lto_mode" == "full" ]]; then
+      scripts/config --set-str "DRM_PANIC_SCREEN" "kmsg"
+      _disable "DRM_PANIC_SCREEN_QR_CODE"
+      _undefine "DRM_PANIC_SCREEN_QR_CODE_URL" "DRM_PANIC_SCREEN_QR_VERSION"
+    fi
   fi
 
   if [ "$_distro" = "Gentoo" ]; then


AdelKS:
Do you have a kernel with this on, can you test it by triggering a panic hehe ? I bet you didn't know you can artifically trigger a panic, or at least check it's working (although even if it doesn't work it's would be as bad as current state, but you know, better have something better than replace it with something as bad hehe). Don't know how to make sure you don't break anything on disc, probably remount read only first.

damachine:
> Do you have a kernel with this on, can you test

currently i have a `no` lto build i wanted test if rust build well, but i could test this, yeah it makes no sense if it doesn't work and jumps to `user` again. 

> I bet you didn't know you can artifically trigger a panic

hehe you are right i did not know it, never needed that, ha.

Okay, i look into it, I will give you an update if it works

damachine:
I do it tomorrow must build the kernel for this hehe

okay thanks llm and what i read here https://docs.kernel.org/admin-guide/sysrq.html#what-are-the-command-keys

do you think this is mostly safe or i miss something
```
sync
echo s | sudo tee /proc/sysrq-trigger # sync all mounted filesystems
echo u | sudo tee /proc/sysrq-trigger # remount all mounted filesystems read-only
echo c | sudo tee /proc/sysrq-trigger # system crash O)
```

all from the `tty` of course

AdelKS:
> do you think this is mostly safe or i miss something

Yeah that's perfect, sync all disk ops, remout ro, now your disk is safe, can go nuclear

damachine:
It seems Nvidia doesn't support the DRM `panic_screen` feat. hehe.

Neither

```
❯ cat /sys/module/drm/parameters/panic_screen
qr_code
```

nor

```
❯ cat /sys/module/drm/parameters/panic_screen
kmsg
```

shows for me the `panic_screen` if I go nuclear.
I don't think I have anything exotic that blocking it.
It seems this hasn't been resolved for Nvidia yet, based on what I've found from 2024 o) https://forums.developer.nvidia.com/t/drm-panic-support/314435

AdelKS:
Alright, time for me to try it then

AdelKS:
Looks like it's working,  now my PC doesn't work anymore and you owe me an upcoming RDNA5 GPU 😌

AdelKS:
Thanks for this little improvement! Merging.

damachine:
> Looks like it's working, now my PC doesn't work anymore and you owe me an upcoming RDNA5 GPU 😌

cool, guess NVIDIA will deny me that pleasure :smiley:

PR #1224, opened 2026-07-16
config: move OpenRGB support to legacy

damachine:
OpenRGB kernel patch is deprecated. Its AMD SMBus support has been upstream since Linux 5.8, while the remaining Nuvoton SMBus driver is maintained separately as i2c-nct6793-dkms. 
https://gitlab.com/OpenRGBDevelopers/OpenRGB-Wiki/-/blob/stable/OpenRGB-Kernel-Patch.md
https://gitlab.com/CalcProgrammer1/OpenRGB/-/work_items/4221#note_2048005846

I suggest moving the patch to the legacy options and keeping it available for NCT6775/6776 hardware, while disabling it by default. Also fix the inconsistent configuration where `I2C_NCT6775` was enabled even when the OpenRGB patch was disabled.

Maybe it can be removed once it no longer has a supported use case.

### DIFF
--- a/customization.cfg
+++ b/customization.cfg
@@ -240,9 +240,6 @@ _custom_commandline="intel_pstate=passive"
 # Selection of Clearlinux patches
 _clear_patches="true"
 
-# Add OpenRGB compatibility for certain i2c controllers - https://gitlab.com/CalcProgrammer1/OpenRGB/-/blob/master/OpenRGB.patch
-_openrgb="true"
-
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Yeah that was the exact reason we kept it in for that long. I guess we can leave it be as legacy indeed.

Tk-Glitch:
Thanks damachine! Let's merge.
:frog: :heart:

PR #1225, opened 2026-07-22
 _libunwind_replace fails with Gentoo default libgcc runtime

damachine:
Hey 🐸 ❤️

Unless I missed something, what is the intended configuration for `_libunwind_replace`?

Since linux-tkg supports this option, I decided to use it to make the build process more consistent when primarily using LLVM toolchains :D

To do this, I enabled the `llvm-libunwind` USE flag 

```
llvm-core/clang-common llvm-libunwind
llvm-runtimes/clang-runtime llvm-libunwind polly
llvm-runtimes/clang-unwindlib-config llvm-libunwind
```

and set `_libunwind_replace=true`, but Clang still defaults to using `libgcc`.

```text
...
 -> Using modprobed-db
make --no-print-directory -C /DATA/damachine/Projekte/linux-tkg/linux-src-git \
-f /DATA/damachine/Projekte/linux-tkg/linux-src-git/Makefile localmodconfig
make -f ./scripts/Makefile.build obj=scripts/basic
# HOSTCC  scripts/basic/fixdep
  clang -Wp,-MMD,scripts/basic/.fixdep.d -Wall -Wmissing-prototypes -Wstrict-prototypes -O2 -fomit-frame-pointer -std=gnu11   -I ./scripts/include    -unwindlib=libunwind -o scripts/basic/fixdep scripts/basic/fixdep.c   
clang: error: --rtlib=libgcc requires --unwindlib=libgcc
make[2]: *** [scripts/Makefile.host:114: scripts/basic/fixdep] Fehler 1
make[1]: *** [/DATA/damachine/Projekte/linux-tkg/linux-src-git/Makefile:666: scripts_basic] Fehler 2
make: *** [Makefile:248: __sub-make] Fehler 2
 -> exit cleanup done
```

Pairing it with `-rtlib=compiler-rt` in `HOSTLDFLAGS` fixes the `HOSTCC` failure:

```bash
export HOSTLDFLAGS="-rtlib=compiler-rt -unwindlib=libunwind"
```

Enabling Gentoo s `default-compiler-rt` USE flag would also work, but it changes the default for every Clang invocation in that slot and makes `_libunwind_replace` redundant. The toggle should select the compatible runtime pair itself.

And I think `LDFLAGS_MODULE` should also be removed. The kernel and its modules do not link against these userspace runtimes. Since `-rtlib` and `-unwindlib` are Clang driver options, they belong in `HOSTLDFLAGS`, not in flags passed directly to `$(LD)` when linking modules.

---

Found one more small thing in the install flow, so I added it here too.

The prompts now show `[N]/y` to make the default clear. On Gentoo, `make install` now runs after `@module-rebuild`, so out-of-tree modules are available before any configured installkernel hooks regenerate the initramfs or UKI.

Cheers

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -214,7 +214,7 @@ if [ "$1" = "install" ]; then
     fi
 
     if [[ "$_install_after_building" = "prompt" ]]; then
-      read -p "Do you want to install the new Kernel ? Y/[n]: " _install
+      read -p "Do you want to install the new Kernel ? [N]/y: " _install
     fi
 
     if [[ "$_install_after_building" =~ ^(Y|y|Yes|yes)$ || "$_install" =~ ^(Y|y|Yes|yes)$ ]]; then
@@ -265,7 +265,7 @@ if [ "$1" = "install" ]; then
     fi
 
     if [[ "$_install_after_building" = "prompt" ]]; then
-      read -p "Do you want to install the new Kernel ? Y/[n]: " _install
+      read -p "Do you want to install the new Kernel ? [N]/y: " _install
     fi
 
     if [[ "$_install_after_building" =~ ^(Y|y|Yes|yes)$ || "$_install" =~ ^(Y|y|Yes|yes)$ ]]; then
@@ -294,7 +294,7 @@ if [ "$1" = "install" ]; then
         warning "By default, system kernel updates will overwrite your custom kernel."
         warning "Adding a lock will prevent this but skip system kernel updates."
         msg2 "You can remove the lock if needed with 'sudo zypper removelock kernel-default-devel kernel-default kernel-devel kernel-syms'"
-        read -p "Would you like to lock system kernel packages ? Y/[n]: " _lock
+        read -p "Would you like to lock system kernel packages ? [N]/y: " _lock
         if [[ "$_lock" =~ ^(Y|y|Yes|yes)$ ]]; then
           sudo zypper addlock kernel-default-devel kernel-default kernel-devel kernel-syms
         fi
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -392,23 +392,23 @@ if [ "$1" = "install" ]; then
...
         sudo emerge @module-rebuild --keep-going
       fi
     fi
 
+    msg2 "Installing kernel"
+    sudo make install
+
+    sudo cp "$(make ${llvm_opt} -s image_name)" "/lib/modules/$_kernelname/vmlinuz"
+
   fi
```
AdelKS: I'd rather keep this before `emerge @module-rebuild` in case it fails, the install is complete.

damachine: Hey Adel, quick question about the Gentoo module rebuild flow.

`install.sh` uses `set -e`, so I am considering wrapping the optional rebuild like this:

```sh
if ! sudo emerge @module-rebuild --keep-going; then
  warning "Some external kernel modules failed to rebuild; continuing with kernel installation."
fi
```

My understanding is that `emerge --keep-going` still returns a non-zero exit status if one or more module rebuilds fail, but it first continues with any independent rebuilds it can complete.

Because emerge runs inside `if ! ...`, `set -e` will not abort the script. It will show the warning and continue with `make install`, using whichever modules rebuilt successfully. Or am I overlooking something here?

AdelKS: it's fine, but why you really want to install the kernel after the module rebuild ? actually those module end up in `/lib/modules` isn't it ? wouldn't that folder not exist if we don't do the install first ?

damachine: On my setup, make install triggers the UKI hook. If it runs before the external modules are installed, they won’t be included, so I’d need another hook to rebuild the UKI afterward.

AdelKS: ah! then sounds good, please add a small comment about it in the code and in the commit message then! split it from the y/n fix if possible. Thanks!

AdelKS: so we have `make modules_install` done before that, good, please move the kernel copy to `/lib/modules` at that level too

AdelKS:
Awesome thanks!

PR #1227, opened 2026-07-26
PKGBUILD: provide NTSYNC-MODULE

damachine:
Hey sry again but - the latest CachyOS Wine update exposed this through its ntsync dependency. 

Fix: native NTsync only creates `ntsync.conf`, so `NTSYNC-MODULE` was missing. Use it for the module provider and keep `ntsync.rules` for the legacy header. This also avoids duplicating the provider list. cheers ;)

Bildschirmfoto_20260726_040748


AdelKS:
Thanks!

PR #1228, opened 2026-07-27
Bore Patch Fix for 7.1.5

azurkytheturkey:
patch was broken, hunk 19 wouldnt build. issue is fixed.


AdelKS:
Hey please open this PR [upstream](https://github.com/firelzrd/bore-scheduler), because we sync the patches automatically from there and this fix would be overwriten by the older one.

PR #1233, opened 2026-08-07
ci: fix LTS config sync

damachine:
Hey! `update-kernel-configs.yml` needs a little refresh. <3

https://github.com/Frogging-Family/linux-tkg/actions/workflows/update-kernel-configs.yml


AdelKS:
Thanks ! So now LTS follows non-LTS naming, that was bound to happen hehe

PR #1234, opened 2026-08-09
prepare: Simplify "exit_cleanup" compiler version logging

damachine:
Hi guys <3, hope you’re all doing well!

If i see right on a Gentoo clang default settings, calling `clang -v` inside `exit_cleanup` with `set -e` can cause Clang to link an empty program and exit with code 1.

Im not sure why `install.sh` doesn’t show this, but it could still break things we may have overlooked, such as CI. I noticed it while using tkginstaller, as it reported an error after the installation had already completed. At first, I thought tkginstaller itself was the problem, hehe.

I looked for and tested a cleaner approach and I think this does exactly what we need, it makes the check more robust and avoids a possible exit code 1. `-x c` treats `/dev/null` as empty C input while `-###` prints the commands without running them. What do you think? Cheers :)

Bildschirmfoto_20260809_222305

### DIFF
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -2047,10 +2047,10 @@ exit_cleanup() {
   cat /etc/os-release > "$_where"/logs/system-info.txt
   if command -v clang &> /dev/null; then
     echo "#################" >> "$_where"/logs/system-info.txt
-    clang -v >> "$_where"/logs/system-info.txt 2>&1
+    clang --version 2>&1 | head -n1 >> "$_where"/logs/system-info.txt
   fi
   echo "#################" >> "$_where"/logs/system-info.txt
-  gcc -v >> "$_where"/logs/system-info.txt 2>&1
+  gcc --version 2>&1 | head -n1 >> "$_where"/logs/system-info.txt
 }
 
 trap exit_cleanup EXIT


AdelKS:
Hehe I fixed it recently in my PR #1127, I just appended a `|| true`. I knew about this issue all along, I guess it's worth a gentoo bug, I'll ping people in IRC to ask 🤔 because other distros don't have that weird clang behavior

damachine:
> PR #1127, I just appended a || true

Ah, cool.

you’re right, this is not on Arch. I looked into it again, Looks So this is actually a known Clang issue. xD

https://github.com/llvm/llvm-project/issues/67209

AdelKS:
I think let's go with `clang -v 2>&1 | grep version`, that's all we ever looked at actually

AdelKS:
I guess we can do the same with `gcc -v 2>&1 | grep "gcc version"`

AdelKS:
it seems that `clang -v -###` is enough for it to return a zero exit code

damachine:
ahh for `gcc -v 2>&1 | grep "gcc version"` i need `LC_ALL=C` as non non english locales xD

Suggestion: use `clang/gcc --version | head -n1` ?

AdelKS:
> Suggestion: use clang/gcc --version | head -n1 ?

For me `gcc` shows the version at the end, `clang` at beginning, up to you if you want to use `tail`/`head`, don't forget the `2>&1`

damachine:
> it seems that clang -v -### is enough for it to return a zero exit code

Ah, I think that 's what you mean? We are just changing the `clang` thing. And everything else stays as it is.

AdelKS:
No no, let's go with only saving the version, it was just a side note, sorry x)

damachine:
We might need to revise `system-info.txt` at some point, though there is some duplication, and additional info might be useful. But nice for now we have exactly what we wanted.

Bildschirmfoto_20260810_032341

AdelKS:
Need to reword the commit

AdelKS:
thanks!

damachine:
Sorry :nerd_face:, maybe this is interesting to you as a fellow Gentoo user hehe.

As described in https://github.com/llvm/llvm-project/issues/67209

The issue comes from Gentoos automatically loaded Clang config in `/etc/clang/gentoo-hardened-ld.cfg`:

```
-Wl,-z,relro
-Wl,-z,now
```

Clang treats these linker options as a reason to link, even without an input file. So clang tries to build an empty program.

Gentoo already knows about the bug and it seems to be a migration issue since the `$` syntax  is only supported from LLVM 20 onward.

From: https://clang.llvm.org/docs/UsersManual.html#configuration-files

Bildschirmfoto_20260810_230602

I tested this in `/etc/clang/gentoo-hardened-ld.cfg`.

~~~diff
--Wl,-z,relro
--Wl,-z,now
+$-Wl,-z,relro
+$-Wl,-z,now
~~~

Result:

Bildschirmfoto_20260810_221017

PR #1241, opened 2026-08-21
pkgbuild: drop "record" - install.sh: use time for make

damachine:
Howdy! I hope I'm not stepping on anyone's toes if someone was already planning to tackle this ;) but we noticed that makepkg wasn't capturing the build output, so here's my hopfully KISS proposal to keep the build output visible while logging it to `shell-output.log`.

The previous `return 0` could mask a failed build and allow `makepkg` to continue. Was that needed for something i miss(CI)? or can it be dropped? See ya!

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -124,7 +124,7 @@ build() {
     _schedtool="command schedtool -B -n 1"
     _ionice="command ionice -n 1"
   fi
-  _runtime=$(
+  (
     if [ -n "$_schedtool" ]; then
       _pid="$(exec bash -c 'echo "$PPID"')"
       $_schedtool "$_pid" ||:
@@ -135,8 +135,7 @@ build() {
     export KCFLAGS
     export KRUSTFLAGS
 
-    time ( make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules 2>&1 ) 3>&1 1>&2 2>&3
-    return 0
+    time make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules
   )
 }
 
--- a/install.sh
+++ b/install.sh
@@ -196,7 +196,7 @@ if [ "$1" = "install" ]; then
   if [[ "$_distro" =~ ^(Ubuntu|Debian)$ ]]; then
 
     msg2 "Building kernel DEB packages"
-    make ${llvm_opt} -j ${_thread_num} bindeb-pkg LOCALVERSION=-${_kernel_flavor}
+    time make ${llvm_opt} -j ${_thread_num} bindeb-pkg LOCALVERSION=-${_kernel_flavor}
     msg2 "Building successfully finished!"
 
     # Create DEBS folder if it doesn't exist
@@ -247,7 +247,7 @@ if [ "$1" = "install" ]; then
 
     msg2 "Building kernel RPM packages"
     export RPMOPTS="--define '_topdir ${_fedora_work_dir}' --define 'install_mod_strip 1'"
-    make ${llvm_opt} -j ${_thread_num} binrpm-pkg LOCALVERSION="${_extra_ver_str}"
+    time make ${llvm_opt} -j ${_thread_num} binrpm-pkg LOCALVERSION="${_extra_ver_str}"
     msg2 "Building successfully finished!"
 
     # Create RPMS folder if it doesn't exist
@@ -321,7 +321,7 @@ if [ "$1" = "install" ]; then
     fi
 
     msg2 "Building kernel"
-    make ${llvm_opt} -j ${_thread_num}
+    time make ${llvm_opt} -j ${_thread_num}
     msg2 "Build successful"
 
     if [ "$_STRIP" = "true" ]; then
--- a/linux-tkg-config/prepare
+++ b/linux-tkg-config/prepare
@@ -2053,9 +2053,6 @@ exit_cleanup() {
   fi
 
   msg2 'exit cleanup done\n'
-  if [ -n "$_runtime" ]; then
-    msg2 "compilation time : \n$_runtime"
-  fi
 
   [[ -f "$_where"/git_work.log.txt ]] && \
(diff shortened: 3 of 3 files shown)


PKGBUILD
```diff
@@ -135,8 +135,10 @@ build() {
     export KCFLAGS
     export KRUSTFLAGS
 
-    time ( make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules 2>&1 ) 3>&1 1>&2 2>&3
-    return 0
+    time (
+      make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules 2>&1 | tee "$_where/shell-output.log"
```
AdelKS: You are right, now that we save BIG_UGLY_FROGMINER in the logs, we don't really need to save the "full" console output including user choices anymore 🤔

PKGBUILD
```diff
@@ -135,8 +135,10 @@ build() {
...
+      exit "${PIPESTATUS[0]}"
```
AdelKS: pretty sure it's not `exit` we want here. There must be a documented way to use `| tee` and still return the status of the command before the pipe 🤔

PKGBUILD
AdelKS: call it simply `build.log` ? Yeah I think we can take this route. In the `install.sh` approach we should do the same: drop the use of `script` entirely and just save `build.log`

PKGBUILD
```diff
@@ -135,8 +135,10 @@ build() {
...
+    ) 3>&1 1>&2 2>&3
```
AdelKS: this weirdness is probably vestigial and we can go back to simple `time make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules` ?

PKGBUILD
```diff
@@ -135,8 +135,10 @@ build() {
     export KCFLAGS
     export KRUSTFLAGS
 
-    time ( make ${_force_all_threads} ${llvm_opt} LOCALVERSION= bzImage modules 2>&1 ) 3>&1 1>&2 2>&3
-    return 0
```
AdelKS, line 139: Yeah this `return 0` is weird, so it never fails ? we would have had a lot of issue reports with broken kernels ? weird...

damachine:
The fd swapping only captures the output from `time` in `_runtime`, which is printed later by `exit_cleanup`. If that is no longer useful, then yes, we can simplify it. We would still need `pipefail` so `tee` does not mask a failure from make.
```
(
  set -o pipefail
  time make ... 2>&1 | tee "$_where/shell-output.log"
)
```
Is that what you had in mind?

AdelKS:
is this `_runtime` variable something special in `PKGBUILD`s ? If not, yeah we can even drop having the code in a subshell

damachine:
> is this `_runtime` variable something special in `PKGBUILD`s ? If not, yeah we can even drop having the code in a subshell

Nope, it’s linux-tkg specific and only used to print the build time during `exit_cleanup`. We can drop that too, hehe 😄

AdelKS:
Alright, now I understand, somehow I searched for it wrong and didn't see it in the `exit_cleanup()` function, @Tk-Glitch good with some KISSification ?

customization.cfg
```diff
@@ -273,9 +273,6 @@ _NR_CPUS_value=""
 # Options are: "yes", "no", "prompt"
 _install_after_building="prompt"
 
-# [non-Arch only] Use 'script' command for logging
-_logging_use_script="yes"
```
AdelKS, line 277: add it to the deprecated variables in `prepare` too please

damachine: done ;)

install.sh
```diff
@@ -35,18 +35,20 @@ plain() {
...
+_run_build() (
+  if [[ "$CI" == "true" ]]; then
+    "$@"
+    exit
+  fi
+
+  set -o pipefail
+  "$@" 2>&1 | tee "$_where/logs/build.log"
+)
+
```
AdelKS: I don't think this function is useful, better just have the `2>&1 | tee "$_where/logs/build.log"` at the call site, the `$CI` thing is not useful and `$CI` seems to also not be defined anywhere. Let the CI take exactly the same code path, the only thing we enforce there is `_nofallback=true`

AdelKS:
Please try to think on how split your changes into smaller logical changes that leave the repo usable at each step (note that LLMs are bad at this...), for example here:
- PKGBUILD: drop recording build time in variable
- PKGBUILD: save build output to logs
- install.sh: save build output to logs
- install.sh: stop using `script`

install.sh
damachine: > $CI seems to also not be defined anywhere

Ah, I found this. GitHub always sets CI=true when a workflow is running, so it doesn’t need to be defined explicitly.
https://docs.github.com/en/actions/reference/workflows-and-actions/variables

damachine:
> Please try to think on how split your changes into smaller logical changes that leave the repo usable at each step

Before I split this later into bisect friendly commits xD, one question. Would you still prefer dropping `_run_build()` and adding tee to every make call? I kept the helper because Actions already captures the output, so CI doesn’t really need a separate log file.

AdelKS:
> CI doesn’t really need a separate log file.

It doesn't indeed, but it's not worth creating a separate code path for it: 
1. there will be a useless log file in the CI runner and we simply won't care about it. 
2. In general, it's better to have the CI use the same code path as any user to catch issues

With that removed, `_run_build()` just makes it harder to read to human since you need to look for the function to just see that it adds a `| tee ` to the command.

damachine:
Do we want to use `time` in `install.sh` as well?

Bildschirmfoto_20260824_014825

damachine:
Hey @AdelKS I've reverted your concerns about "script" logging so everything is now logged again.

Let's see, maybe we should still take that

- drop the arch specific `runtime` function. `time` now shows the built time.
- and we should get the weird `return 0` out of here.

AdelKS:
Thanks!

PR #1244, opened 2026-08-30
pkgbuild: package Rust artifacts and unstripped VDSO

### LINKED ISSUE #1243: RUST support for bcachefs
### Same-issue check

Yes

### Linux Distribution

Archlinux

### Kernel version

6.18.48, 7.1.12, 7.2.2

### What happened?

I am testing bcachefs and its capabilities. bcachefs needs CONFIG_RUST=y otherwise it is giving a warning on the commandline when using it. See this issue: https://github.com/Frogging-Family/linux-tkg/issues/1217

What I found today is a bcachefs message in the journal saying:

```
Aug 30 14:58:22 rakete kernel: starting version per_dev_fragmentation_lru (1.39)
                                 with options: btree_cache_shrinker_seeks=4
                                 allowing incompatible features up to need_discard_by_journal_seq (1.38)
                                 with devices: sde,sdf,sdg,sdh
                                 built without Rust support; this will be required in the near future
                                   reason: missing the kernels prebuilt Rust stdlib (/usr/lib/modules/6.18.48-273.2-linux618-tkg-x86-64-v3-300hz/build/rust/libcore.rmeta); the kernel was built/installed without its rust/ artifacts
                                   rebuild the module with rustc, bindgen and rust-src available to enable it
```

The rust config options are all set from my point of view:

```
/lib/modules/6.18.48-1-lts/build/.config :
CONFIG_RUSTC_VERSION=109800
CONFIG_RUST_IS_AVAILABLE=y
CONFIG_RUSTC_LLVM_VERSION=220108
CONFIG_RUSTC_HAS_COERCE_POINTEE=y
CONFIG_RUSTC_HAS_SPAN_FILE=y
CONFIG_RUSTC_HAS_UNNECESSARY_TRANSMUTES=y
CONFIG_RUSTC_HAS_FILE_WITH_NUL=y
CONFIG_RUSTC_HAS_FILE_AS_C_STR=y
CONFIG_RUSTC_HAS_SUSPICIOUS_RUNTIME_SYMBOL_DEFINITIONS=y
CONFIG_RUST=y
CONFIG_RUSTC_VERSION_TEXT="rustc 1.98.0 (88d9e12ae 2026-08-18) (Arch Linux rust 1:1.98.0-1)"
CONFIG_HAVE_RUST=y
CONFIG_RUST_FW_LOADER_ABSTRACTIONS=y
# CONFIG_BLK_DEV_RUST_NULL is not set
CONFIG_ANDROID_BINDER_IPC_RUST=y
# CONFIG_RUST_BITMAP_HARDENED is not set
# CONFIG_FIND_BIT_BENCHMARK_RUST is not set
# CONFIG_RUST_DEBUG_ASSERTIONS is not set
CONFIG_RUST_OVERFLOW_CHECKS=y
# CONFIG_RUST_BUILD_ASSERT_ALLOW is not set
```

I am wondering what I need to do with the linux-tkg kernel to be fully compliant with bcachefs. I have no clue what this journal message is asking me to do.

### Relevant logs

n/a

### My Favorite type(s) of frog

Frog

damachine:
Howdy all <3

`CONFIG_RUST=y` was doing its job, but `hackheaders()` quietly left the generated Rust bits behind. This meant external Rust modules couldn't find required prebuilt artifacts such as `rust/libcore.rmeta`.

Package the generated `.rmeta` and `.so` files when Rust is enabled, matching [Arch Linux's kernel packaging](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/PKGBUILD#L214).

Fixes #1243 

:frog: :crab:

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -273,6 +273,12 @@ hackheaders() {
   msg2 "Installing KConfig files..."
   find . -name 'Kconfig*' -exec install -Dm644 {} "$builddir/{}" \;
 
+  if [[ $(scripts/config -s CONFIG_RUST) = y ]]; then
+    msg2 "Installing Rust files..."
+    install -Dt "$builddir/rust" -m644 rust/*.rmeta
+    install -Dt "$builddir/rust" rust/*.so
+  fi
+
   msg2 "Removing unneeded architectures..."
   local arch
   for arch in "$builddir"/arch/*/; do
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -279,6 +279,10 @@ hackheaders() {
     install -Dt "$builddir/rust" rust/*.so
   fi
 
+  msg2 "Installing unstripped VDSO..."
+  make INSTALL_MOD_PATH="$pkgdir/usr" vdso_install \
+    link=  # Suppress build-id symlinks
+
   msg2 "Removing unneeded architectures..."
   local arch
   for arch in "$builddir"/arch/*/; do


damachine:
I also noticed we aren’t installing the unstripped vDSO files like the [official Arch Linux PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/PKGBUILD#L220). Would it make sense to add that too? :P

I wasn’t familiar with vDSO, but [`Arch man(7)`](https://man.archlinux.org/man/vdso.7.en) explains it nicely <3

AdelKS:
> I also noticed we aren’t installing the unstripped vDSO files like the [official Arch Linux PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/PKGBUILD#L220). Would it make sense to add that too? :P

No idea, by reading the docs you pointed to, would be weird if it weren't working at all in `linux-tkg` as it sounds like one can have serious perf regressions without it :thinking:

damachine:
> No idea, by reading the docs you pointed to, would be weird if it weren't working at all in `linux-tkg` as it sounds like one can have serious perf regressions without it 🤔

Yeah, the link I posted was too general. You’re right vDSO itself works fine in linux-tkg, and the unstripped files are mainly useful for debugging and profiling. So adding them is just a nice packaging addition rather than a runtime fix. :P

PKGBUILD
```diff
@@ -273,6 +273,16 @@ hackheaders() {
   msg2 "Installing KConfig files..."
   find . -name 'Kconfig*' -exec install -Dm644 {} "$builddir/{}" \;
 
+  if [[ $(scripts/config -s CONFIG_RUST) = y ]]; then
```
AdelKS, line 276: is it better to copy if the folder exists instead ?

PKGBUILD
```diff
@@ -273,6 +273,16 @@ hackheaders() {
...
+    msg2 "Installing Rust files..."
+    install -Dt "$builddir/rust" -m644 rust/*.rmeta
+    install -Dt "$builddir/rust" rust/*.so
```
AdelKS, line 279: Does `make install` cover it in our `install.sh` approach ? Sorry for dropping this on you x)

PKGBUILD
```diff
@@ -273,6 +273,16 @@ hackheaders() {
...
+  fi
+
+  msg2 "Installing unstripped VDSO..."
+  make INSTALL_MOD_PATH="$pkgdir/usr" vdso_install \
+    link=  # Suppress build-id symlinks
+
```
AdelKS, line 285: Does `make install` cover this ?

Oh, this one installs to `/usr`, what's tree under, `/usr/src/linux` ? that can be more touchy

PKGBUILD
damachine: the rust dir can also exist when CONFIG_RUST is disabled, so I think the config check is safer. Older kernels should handle an unknown CONFIG_RUST cleanly and return undef

PKGBUILD
damachine: No worries x) make install doesn’t cover it, but install.sh copies the entire build tree to `/usr/src` first, so the Rust artifacts are already included there

DEB/RPM use the upstream paths, and their current header packaging doesn’t seem to copy the Rust artifacts either.

AdelKS: okay so we consider this properly handled in "Generic" and "Gentoo", for the other distros, fingers crossed haha

PKGBUILD
AdelKS: same answer as you gave me above, we good.

PKGBUILD
AdelKS: > the rust dir can also exist when CONFIG_RUST is disabled

Yeah that's the thing, and you are saying it's not useful at that moment ? I'd rather have a more stupid test that just add few useless files than start tapping into CONFIG options that change every sunday

damachine: > Yeah that's the thing, and you are saying it's not useful at that moment ?

Right, the dir exists but without CONFIG_RUST the `.rmeta` and `.so` files aren’t built, so there’s nothing to copy. Or did I misunderstand, did you mean checking for the actual files as an extra safeguard?

AdelKS: I meant we could blanket copy that folder if it exists, no matter what's in it if doing so never breaks anything

damachine: Do you mean copying the entire rust/ directory, or just installing the .rmeta and .so files whenever they exist?

damachine: Ah, okay sorry xD, now I get what you mean. If that amount of ballast is acceptable, I’m fine with the simpler approach. We could copy the entire dir, it’s only around 7 MiB currently and yeah the files should not break anything. 

My only concern is that it may grow over time, and most of those files aren’t needed in the headers package. That’s probably why Arch only installs the generated artifacts, which seems more futureproof.

damachine: Sorry for the confused back and forth earlier I needed a moment to think it through properly hehe.

I get the KISS angle, and I understand the concern about Kconfig changing every Sunday. The list of required Rust artifacts may grow over time, but I don’t think `CONFIG_RUST` itself is going to change every Sunday at 2-minute intervals lol.

My concern with using `rust/` as the signal is that the directory only tells us the kernel sources support Rust, not that this particular kernel was built with it. With `CONFIG_RUST=n`, those files aren’t useful for external Rust modules, and a reused build tree could even contain stale Rust artifacts from an earlier build.

`CONFIG_RUST` has been the top-level switch since Rust entered mainline, `scripts/config` returns `undef` on kernels that don’t know it, and Arch currently uses this exact check as well. So checking the final config and installing only the generated `.rmeta` and `.so` files still seems like the simplest semantic test to me. But I’m fine with your preference if you’d rather blanket copy the directory. Fair nerdy concerns on both sides <3

AdelKS: sounds good! so we go with the config check ! Plus people will insist in the "disable rust" toggle eventually and won't want to have that folder included. So yeah let's go with copying the minimal amount of things from that folder as you said.

AdelKS:
Thanks! <3

PR #1247, opened 2026-09-07
config: update modprobed-db path and mark Linux 7.1 EOL

damachine:
Howdy <3

Just a tiny housekeeping PR.

This updates the default modprobed-db path. Upstream moved the database from `~/.config/modprobed.db` to `~/.local/share/modprobed-db/modprobed.db` in [cbe8b0d](https://github.com/graysky2/modprobed-db/commit/cbe8b0dbb415e1b5fb2d9c85da0242fa8d239272).

I also figured we could go ahead and move Linux 7.1 to the EOL list.

~~Nothing wild.~~ ;) I'm always happy to contribute.

Update:

Ah, I think I was a bit too quick with modprobe-db. It has been updated on Arch, but perhaps not yet on all distros, so the old path is still the default. So, let's hold off on that for a bit. But everything is ready. o)


AdelKS:
> So, let's hold off on that for a bit. But everything is ready.

You are saying that users who already have a db, get it moved by simply updating ? With some post install script within the package ?

damachine:
Hi! There’s no post install script. Version 2.51 migrates an existing .db from the old default `DBPATH` to the new one the next time modprobed-db runs. My concern is that some distros still ship older versions and therefore still use the old path. 

I’m not sure how we should handle it. Since users can already customize the path. 

Maybe a short note like this is enough, and we use the new path by default. What do you think? :P

```sh
# modprobed-db database file location (v2.51+; older versions use ~/.config/modprobed.db)
_modprobeddb_db_path=~/.local/share/modprobed-db/modprobed.db
```

damachine:
We could also leave `_modprobeddb_db_path` empty and let linux-tkg check both the new and old default paths automatically. A custom path would still take priority. I looked into it, and the changes would be minimal hehe.

linux-tkg-config/prepare
```diff
@@ -1853,18 +1853,26 @@ fi
 
   if [[ "$_modprobeddb" = "true" || "$_kernel_on_diet" == "true" ]]; then
     msg2 "Using modprobed-db"
-    if [ -f "$_where"/"$_modprobeddb_db_path" ]; then
+    if [[ -n "$_modprobeddb_db_path" && -f "$_where/$_modprobeddb_db_path" ]]; then
       _modprobeddb_db_path="$_where"/"$_modprobeddb_db_path"
     fi
+    if [[ -z "$_modprobeddb_db_path" || ! -f "$_modprobeddb_db_path" ]]; then
```
AdelKS: keep only `if [[ -z "$_modprobeddb_db_path" ]]` so we fail when it's filled with a wrong path

AdelKS:
Thanks!

Issue #1248, opened 2026-09-19
Disable numa: warning: symbol value 'n' invalid for NODES_SHIFT


cheako:
### Same-issue check

Yes

### Linux Distribution

Debian

### Kernel version

7.3.rc3

### What happened?

My autism means I won't try and read your mind, I just cannot think of anything else you'd like to know.  It's not that I just left the report incomplete.  Please reply assuming answers to yes/no questions, I.E. include what to do given each branch.

Thanks!

This one file should be enough to reproduce the build, feel free to share it.  It was built with input from Gemini, for example I suggested using `find` but I had the LLM generate the specific arguments used.
[Dockerfile.linux-tkg.txt](https://github.com/user-attachments/files/32419478/Dockerfile.linux-tkg.txt)
[console.log](https://github.com/user-attachments/files/32419466/console.log)

It may be helpful to know that Gemini completely gave up trying to work through this and so I abandoned trying to disable numa.

### Relevant logs

[system-info.txt](https://github.com/user-attachments/files/32419491/system-info.txt)
[prepare.log.txt](https://github.com/user-attachments/files/32419489/prepare.log.txt)
[customization.cfg.txt](https://github.com/user-attachments/files/32419492/customization.cfg.txt)
[git_work.log.txt](https://github.com/user-attachments/files/32419490/git_work.log.txt)

### My Favorite type(s) of frog

Frog, American Bullfrog (Lithobates catesbeianus)

damachine:
I also found the cause of the `NODES_SHIFT` warning. 

<img width="825" height="87" alt="Image" src="https://github.com/user-attachments/assets/9b517701-0749-4cb2-8e79-d2f39010da82" />

I think it depends on NUMA, and Kconfig removes it automatically when NUMA is disabled, and also`NODES_SHIFT` is an integer and should not be passed to `_disable`. 

https://github.com/Frogging-Family/linux-tkg/blob/5852ffca1f5b8e32821868852cf99d1c155a1f38/linux-tkg-config/prepare#L1519

The fix should simply be to remove `NODES_SHIFT` from that `_disable` list. 🙂

cheako:
I agree, that looked troubling to me as well.  But `_disable` just calls `config` with an option that should result in the option just being commented: https://github.com/torvalds/linux/blob/master/scripts/config#L181 and not setting to value to `n`.  So I think there is more to this.

damachine:
[...]

PR #1250, opened 2026-09-19
prepare: stop disabling NODES_SHIFT

damachine:
Hey guys <3, can you confirm this findings? The bug was reported here https://github.com/Frogging-Family/linux-tkg/issues/1248.

So what I see `NODES_SHIFT` is an integer:

https://github.com/torvalds/linux/blob/v7.3-rc3/arch/x86/Kconfig#L1546-L1552

Disabling it produces an invalid value warning. Kconfig handles it automatically when NUMA is disabled.

KISS


AdelKS:
Thanks! 🐸

